PR #2, opened 2025-04-10
fix: FreeRDP on Arch Linux, feat: Check if FreeRDP is installed on setup

### LINKED ISSUE #1: FreeRDP command breaks WinBoat on Arch Linux
WinBoat uses the command `xfreerdp` to connect to the remote desktop of the VM. This breaks down on Arch Linux as the FreeRDP package there uses the command `xfreerdp3`.

## Expected Behavior

When using apps in WinBoat, they should be opened using FreeRDP.

## Current Behavior

Nothing happens when clicking on an app, because of the discrepancy between the commands.

## Steps to Reproduce

1. Install WinBoat
2. Get the [FreeRDP](https://archlinux.org/packages/extra/x86_64/freerdp/) package from the AUR
3. Click on one of the apps in WinBoat

## Context (Environment)

Neofetch system information: http://0x0.st/8_l7.txt
Kernel information: `Linux levev-desktop 6.13.6-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 07 Mar 2025 20:18:44 +0000 x86_64 GNU/Linux`

Levev:
Fixes #1


TibixDev:
LGTM! 🎉

PR #4, opened 2025-04-15
Finish `Sort By: Usage` feature

### LINKED ISSUE #3: Apps carrying the same name multiply themselves in the Apps section
When interacting with the apps such as changing the sorting method or using the search bar, duplicate apps can multiply themselves and quickly take over the entire apps section. Some apps can be valid duplicates, as machines can carry executables of the same program compiled for different architectures.

## Expected Behavior

The number of apps should remain the same regardless of the type of interaction.

## Current Behavior

Any type of interaction that changes the order of apps causes duplicates to multiply

## Steps to Reproduce

1. Install WinBoat
2. Under `Apps`, change the sorting method or use the search bar
3. Watch as duplicate apps take over the entire view

## Possible solutions

Use `WinApp.Path` as the key for the v-for instead of `WinApp.Name` in [`Apps.vue`](https://github.com/TibixDev/winboat/blob/9dfbed4f28185febce37888e61de170068ab5b37/src/renderer/views/Apps.vue#L32)

## Context (Environment)

Neofetch system information: http://0x0.st/8fwF.txt
Kernel information: Linux levev-desktop 6.13.8-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Sun, 23 Mar 2025 17:17:17 +0000 x86_64 GNU/Linux

Levev:
While we're at it, also fixed #3

### DIFF
--- a/src/renderer/lib/winboat.ts
+++ b/src/renderer/lib/winboat.ts
@@ -8,7 +8,6 @@ import PrettyYAML from "json-to-pretty-yaml";
 import { InternalApps } from "../data/internalapps";
 import { getFreeRDP } from "../utils/getFreeRDP";
 import { WinboatConfig } from "./config";
-import { app } from "electron";
 const nodeFetch: typeof import('node-fetch').default = require('node-fetch');
 const path: typeof import('path') = require('path');
 const fs: typeof import('fs') = require('fs');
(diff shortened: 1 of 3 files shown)


TibixDev:
* This also implements caching from what I see

LGTM! 🙌

PR #15, opened 2025-08-30
Migrated to PNPM and removed Vite's chunk size warning

TheDokT0r:
* Migrated to project to pnpm for faster dependencies installation.
* Removed the chunk size warning limit in the Vite config, as there's no need for it in a local project.


TibixDev:
Hey, I'm not against switching to pnpm, but could you show some benchmarks with the differences to see if it'd be worth it?

The Vite ignore chunk size change is a good idea and I'll merge it anyway if you make that a separate PR. :)

orkwitzel:
> Hey, I'm not against switching to pnpm, but could you show some benchmarks with the differences to see if it'd be worth it?
> 
> The Vite ignore chunk size change is a good idea and I'll merge it anyway if you make that a separate PR. :)

It's hard to benchmark pnpm against npm, but because pnpm caches all of your packages on the local machine by default, it makes dependencies installation way faster on local machine, as most people would download the same package for multiple projects.
You can check the benchmarks on Pnpm's [website](https://pnpm.io/benchmarks). Although it's rather bias, it's well known that it's significantly faster then npm.

I'll gladly move the chunk size warning to a new pull request, but I do think the project should switch from vanilla npm (it doesn't have to migrate to pnpm, but there are other amazing projects like BunJS that also speed up the installation process significantly).

ArjixWasTaken:
Even if pnpm was slower (which it isn't), I'd still implore you to switch to it.
pnpm helps prevent the node_modules black hole, you'll never have to redownload the same package+version combo on your computer again, saving storage space

Schaka:
Just here to voice my opinion against npm and for either pnpm or even yarn.
Obviously pnpm would be the quickest plug and play switch - and it's faster in my experience, but more importantly prevents me from bloating my local installations during development.

PR #22, opened 2025-08-31
Removed chunk size warning from vite

TheDokT0r:
As for my previous pull request, this pull request only removed the chunk size warning limit. It's a small change but a nice one imo


TibixDev:
LGTM, thank you! 🙌

PR #43, opened 2025-09-03
feat: add support for adding a custom iso

AdianKozlica:


src/renderer/views/SetupUI.vue
```diff
@@ -487,6 +492,8 @@ const currentStepIdx = ref(0);
 const currentStep = computed(() => steps[currentStepIdx.value]);
 const windowsVersion = ref<WindowsVersionKey>("11");
 const windowsLanguage = ref("English");
+const customIsoPath = ref(undefined);
```
TibixDev: Should be `""` to avoid issues with v-model. `""` evaluates to false anyway when checking if there's any value present.

src/renderer/views/SetupUI.vue
```diff
@@ -487,6 +492,8 @@ const currentStepIdx = ref(0);
...
+const customIsoFileName = ref(undefined);
```
TibixDev: Should be `""` to avoid issues with v-model. `""` evaluates to false anyway when checking if there's any value present.

src/types.ts
```diff
@@ -21,6 +21,7 @@ export type InstallConfiguration = {
     diskSpaceGB: number;
     username: string;
     password: string;
+    customIsoPath?: string | undefined;
```
TibixDev: `value?: T` already evaluates to `T | undefined`, so you can omit `| undefined`

src/renderer/views/SetupUI.vue
```diff
@@ -410,6 +412,9 @@ import { WINDOWS_VERSIONS, WINDOWS_LANGUAGES, type WindowsVersionKey } from "../
 import { InstallManager, type InstallState, InstallStates } from '../lib/install';
 import { openAnchorLink } from '../utils/openLink';
 import license from '../assets/LICENSE.txt?raw'
+
+const path = require('path');
```
TibixDev: Add type information, since CJS modules don't give that to us by default and we can't use `import` for Node modules .

Should be `const module: typeof import('module') = require('module')` syntax.

src/renderer/views/SetupUI.vue
```diff
@@ -410,6 +412,9 @@ import { WINDOWS_VERSIONS, WINDOWS_LANGUAGES, type WindowsVersionKey } from "../
...
+const { dialog } = require('electron').remote || require('@electron/remote');
```
TibixDev: Add type information, since CJS modules don't give that to us by default and we can't use `import` for Node modules .

Should be `const module: typeof import('module') = require('module')` syntax.

AdianKozlica:
Hello, I think I have fixed all the mentioned issues.

The only ones that I haven't touched are
* Displaying an another wallpaper (I don't know which image would be appropriate + it automatically selects the win10 wallpaper if there is no other)
* Setting the language as custom (since the keyword custom is not supported I could theoretically just output "Custom" if the custom iso path is set, but keep the default English)

Btw during my testing, I have noticed that the reset winboat button hangs even if it has deleted the winboat volume and the container

https://github.com/user-attachments/assets/799313e3-c616-4398-aab9-77ee46a8396a


https://github.com/user-attachments/assets/6ca0baf3-38c4-46fd-936f-ad4e8d3156fc

TibixDev:
Thanks for the swift fixes!

A few more small nitpicks then it's ready to merge:
- `Custom:` should say `Custom ISO (Optional)`
- `Selected:` should be semibold

TibixDev:
As for the reset WinBoat button, I suspect its because I call unlink on a directory, which I have found out is not the right approach. It produces an error and the function never finishes. I'll address it in a fix later, or you can try to in a separate PR if you want to.

TibixDev:
Not that, I meant this label
image

TibixDev:
Awesome, thank you for working on this!

PR #55, opened 2025-09-04
add flake.nix

0kenx:
Added flake.nix for NixOS installation.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1756787288,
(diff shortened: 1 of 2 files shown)


TibixDev:
Hi, thanks for your Nix PR!

I don't personally use Nix so I don't exactly know how packaging works there, but would it be possible for it to not statically hardcode 0.7.2 but rather always point to the latest release? This way new releases could be shipped to Nix folks I guess instantly.

0xDracula:
I'm going to test it, and give feedback

0xDracula:
It worked fine for me, but I had to launch winboat from the terminal after running `newgrp docker`, as it doesn't detect that my user is in the docker group. Can't determine if it's a problem related to WinBoat or the flake itself.

But overall experience was fine.

EliasBorrajo:
Yes there os a way to make the version not hardcoded, it should become a variable put into the inputs and then called where it's currently hard coded.

Not an expert in nixos, started recently, but this is feasible.

IceDBorn:
@0kenx I currently wrap the bin with "sg docker" to fix the docker group missing issue.

Kisakay:
the flake have fixed the docker group issue for me ! Amazing!

TibixDev:
I'm still very open to merging this, but could you folks make it dynamic when it comes to versioning so I don't have to keep a separate file updated?

0kenx:
Makes sense, I'll update the config this weekend.

0kenx:
Now dynamic versioning is set up properly. Because nix requires SHA256 of the appimage, a github action calculates and updates the SHA for every release. @TibixDev

0kenx:
Also the docker group issue is fixed.

TibixDev:
I think all concerns I had are fixed now, merging this. 🙌

0xDracula:
#164

IceDBorn:
image

I still have the same error using the appimage, maybe you should wrap it with "sg docker"

PR #58, opened 2025-09-05
Check whether an ISO is a Windows ISO

AdianKozlica:
Using libosinfo we can check whether an ISO belongs to the WinNT OS family.

This is not perfect since some ISOs won't be recognized (e.g. Tiny10), but we can prompt the user telling them it is an Unknown OS and that they should proceed with caution.


TibixDev:
Hi, this is a cool idea but ideally don't involve a fifth programming language, would it be possible to do this in vanilla JS with or without some dependency?

AdianKozlica:
Hello, I have tried solving it using the node-gtk library, but it seems like it doesn't really integrate well with Electron.

We could call the ```osinfo-detect``` command, and filter the output to gets its OS name and to check whether it is bootable or not.

TibixDev:
`osinfo-detect` doesn't seem preinstalled on most distros though, and asking users to install even more packages would kind of hinder the setup experience. I think it should be possible to get something rudimentary done in TS that just scans the ISO and maybe checks for some Windows magic bytes?

AdianKozlica:
I think I solved it.

We can extract the Volume ID from the ISO file, and then match it against the regexes that I have extracted out of the database generated by ```osinfo-db-export --system```.

We can also check whether the ISO is bootable by extracting the boot indicator.

TibixDev:
Hey, thanks for the changes, can you confirm this fully works?

Ideally a little demo video where you select a few Windows ISO-s and show how it detects them, also selecting some Linux ISO-s to show how it rejects them.

I believe the user shouldn't be allowed to continue if a non-Windows ISO is selected or if said ISO is not bootable, and prompts should be added somewhere explaining why. (Without a scrollbar preferably, we wanna fit everything in one screen)

Let me know how you think we should proceed from a UI/UX standpoint. :smile:

AdianKozlica:
Hello, so regarding non Windows ISO files (bootable ones), I don't know what would be the most practical solution. 

We could export all non Windows ISO volume IDs from osinfo db but it would have to be constantly updated, and osinfo itself does not contain every volume ID.

The other option is manually checking ISO files whether they contain a certain file (e.g. for Linux the vmlinuz file), but I don't know how reliable would it be.

Or we could keep it as is and just warn the user that they have selected an unknown OS.

https://github.com/user-attachments/assets/be8578a9-a4ef-4bbc-b4d4-ddf103e22fdb

orkwitzel:
I think that's a great idea! I think it should be merged to main

Stensel8:
Looks cool. Any updates on it?

PR #60, opened 2025-09-05
Add Dynamic USB Passthrough

Levev:
Merging this would mean USB passthrough is finally implemented!


Levev:
Need to rebase branch to main

src/renderer/views/Config.vue
```diff
@@ -81,32 +80,121 @@
...
+                                    </x-card>
+                                </TransitionGroup>
+                                <x-button 
+                                    v-if="availableDevices.length > 0"
+                                    class="mt-4 !bg-gradient-to-tl from-blue-400/20 shadow-md shadow-blue-950/20 to-transparent hover:from-blue-400/30 transition"
+                                    @click="refreshAvailableDevices()"
+                                >
+                                    <x-icon href="#add"></x-icon>
+                                    <x-label>Add Device</x-label>
+                                    <TransitionGroup ref="usbMenu" name="menu" tag="x-menu">
```
waffles-dev: Restricting the height of the list seems to be enough. It already follows overflow-auto rules if using with a number <100.
image
 

```suggestion
                                    <TransitionGroup ref="usbMenu" name="menu" tag="x-menu" class="max-h-64">
```

Levev: Thanks for pointing this out!

Levev:
We could use fuser's `-k` flag to get access to MTP devices on KDE like so:
`$ fuser -k /dev/libmtp-<busnumber>-<portnumber>`
This seems to work, as it kills the hogging kiod6 process (see: https://bugs.kde.org/show_bug.cgi?id=412257)

Although on environments where the MTP handler is based on something else (like GNOME's gvfs-mtp) this may not work, or may not even be necessary, so furher testing is required.

TibixDev:
Finally, LGTM! 🙌
(Here's me hoping nothing blows up)

PR #65, opened 2025-09-07
Add author information to electron-builder config

gliptak:
Targeting to correct https://github.com/TibixDev/winboat/actions/runs/17501386273/job/49714873114


TibixDev:
Hey, I believe this was fixed directly in the next commit e6684bc, but mine targets package.json which still works.

PR #69, opened 2025-09-07
Disallow multiple WinBoat instances.

Levev:
Several things break when WinBoat is opened multiple times, one great example is the QMP connection in #60.
This PR disallows such behaviour using electron's [app.requestSingleInstanceLock()](https://www.electronjs.org/docs/latest/api/app\#apprequestsingleinstancelockadditionaldata), focusing the original instance instead.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -1,4 +1,4 @@
-import { app, BrowserWindow, ipcMain, session } from 'electron';
+import { app, BrowserWindow, ipcMain, session, dialog } from 'electron';
 import { join } from 'path';
 import { initialize, enable } from '@electron/remote/main';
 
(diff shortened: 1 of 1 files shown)


TibixDev:
Hey, this is a great change, didn't know Electron had it built-in. Could we also show some simple alert dialog to the user telling them that WinBoat is already running and that only one instance can run at a time?

Levev:
> Hey, this is a great change, didn't know Electron had it built-in. Could we also show some simple alert dialog to the user telling them that WinBoat is already running and that only one instance can run at a time?

Got it, now I notify the user about this using Electron's [dialog.showMessageBoxSync()](https://www.electronjs.org/docs/latest/api/dialog#dialogshowmessageboxsyncwindow-options)

TibixDev:
Thanks, LGTM! 🙌

PR #92, opened 2025-09-09
made a install line

ghost:


TibixDev:
Hey, I'm generally in favour of this change, but:
- We already have a "Downloading" section which already roughly does what this PR aims to do, but not as a table
- Don't think the distro names need issue links after them, it's generally not clean.
- The commands don't exactly work, you have to scout out the latest release version and edit them, so it's not exactly seamless, wouldn't it be easier to say something like `.rpm in Releases` with a link to Releases?
- The Arch package is not official and that should be indicated
- Why use this method for Arch when AUR helpers can get it done in a single command with pkgbuilds?
- I don't think the Nix package is out yet

This PR still needs tons of polish in my opinion. Also please use better titles in your PR and have a description next time.

ghost:
since aur helpers are not supported by arch, https://wiki.archlinux.org/title/AUR_helpers I did not add yay -S winboat

killertofus:
@TibixDev

PR #117, opened 2025-09-12
Added Password confirmation

Simon-Weij:
Just felt like this was missing in the installer, if you make a typo it can be really annoying


TibixDev:
LGTM! 🙌

PR #118, opened 2025-09-12
Renamed 'cores' to 'threads'

Simon-Weij:
In the installer it lets you select a certain amount of cores, but it actually means the amount of threads. I fixed it :D

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "winboat",
-    "version": "0.7.2",
+    "version": "0.7.10",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "winboat",
-            "version": "0.7.2",
+            "version": "0.7.10",
             "dependencies": {
                 "@electron/remote": "^2.1.2",
                 "@iconify/vue": "^4.3.0",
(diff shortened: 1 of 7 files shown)


TibixDev:
Thanks, is Configuration.vue also affected?

Simon-Weij:
It didn't, thanks for the catch. I now updated it and it should reflect everywhere

TibixDev:
I wonder, would the container support specifying threads instead of cores, have you tested this behaviour? If not, maybe the correction should strive towards fetching cores instead of threads, not the opposite as we have now.

TibixDev:
I wonder, would the container support specifying threads instead of cores, have you tested this behaviour? If not, maybe the correction should strive towards fetching cores instead of threads, not the opposite as we have now.

Simon-Weij:
The code's functionality didn't change, all I've done is renaming cores to threads, it previously said the amount of cores as the amount of threads. I for example have 6 cores, but the program said "12 cores" which actually means the amount of threads. So functionally it works the same. But I just correctly named it "threads"

Simon-Weij:
This current version works fine now, but it isn't backwards-compatiable, so people upgrading will either have to change docker-compose.yml from cores to threads, or fully reinstall.

Levev:
This PR breaks functionality. 

We use [dockur/windows](https://github.com/dockur/windows) for the VM, which is built on top of the [qemu](https://github.com/qemus/qemu) docker image. It expects the field `CPU_CORES` under `environment` to specify the number of "CPU cores" used, as per their [readme](https://github.com/qemus/qemu?tab=readme-ov-file#how-do-i-change-the-amount-of-cpu-or-ram).

Simon-Weij:
Does dockur/windows wrongly say cores instead of threads?

PR #129, opened 2025-09-14
feat: Add `Logs` field to `Bug Report` form

Levev:
Added a field so users can submit their logs by drag & dropping them in the issue form.
Right now it is not required, (See: [L#44](https://github.com/Levev/winboat/blob/20c7e4d051ab49d53a83814eeac347f6df778a0c/.github/ISSUE_TEMPLATE/1-bug-report.yml?plain=1#L44)), although changing that is pretty trivial.


TibixDev:
Thank you, LGTM! 🙌

PR #148, opened 2025-09-18
Fixed an issue where the app list could not be obtained on Japanese Windows 11 Pro.

fukumen:
Hello.

I can't successfully retrieve the app list on a Japanese version of Windows 11 Pro.

When I retrieve the app list from the guest server using the following command, I found that the character encoding is Shift-JIS.

`wget http://localhost:7148/apps`

By changing the character encoding of the apps.ps1 output to UTF-8, I was able to retrieve the app list on a Japanese version of Windows.

### DIFF
--- a/guest_server/main.go
+++ b/guest_server/main.go
@@ -44,14 +44,13 @@ type Metrics struct {
 }
 
 func getApps(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


TibixDev:
Hello, thanks for looking into this, it's a great change, LGTM! 🙌

PR #150, opened 2025-09-18
fix: Remove CPU arguments

kroese:
In the new version (v5.00) of the container, I add this automaticly when an AMD cpu is detected (not an Intel). That removes the need for this ARGUMENT.

P.S. Very cool project!

### DIFF
--- a/src/renderer/lib/install.ts
+++ b/src/renderer/lib/install.ts
@@ -21,7 +21,7 @@ export const DefaultCompose: ComposeConfig = {
     },
     "services": {
         "windows": {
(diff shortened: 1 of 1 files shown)


TibixDev:
Hello, thanks for your PR and a huge thank you for making `dockur/windows`, it's a seriously amazing project and WinBoat would not be possible without it. :heart: 

Could you also change the version of `image` to 5.01?

kroese:
I changed the image version to v5.03

TibixDev:
Thanks, LGTM! 🙌

PR #153, opened 2025-09-19
Chore: Add Prettier workflow

orkwitzel:
Created a new GitHub Action to the pipeline that enforces the Prettier rules before merging to main.
I've also added a few more prettier roles to make it more specific.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,7 @@
 {
   "tabWidth": 4,
   "bracketSameLine": true,
-  "printWidth": 160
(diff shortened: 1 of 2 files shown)


waffles-dev:
I haven't checked with Tibix yet, but this seems useful - can you review why the check is failing please?

.prettierrc
waffles-dev, line 1: Hi - if you'd still like to get this in, please catch up to the latest main (and please use our prettier settings + match the version in prettier.yml to the one in package.json)

PR #154, opened 2025-09-19
Add option to specify #RDP port and error handling if selected port is already in use.

heros-os:


TibixDev:
Hey, thanks for your contribution. Ideally contributions related to port selection should address all ports, including VNC, RDP, GS, QMP.

This issue is already being developed in the `feat/port-selection` branch, however we cannot merge it yet because it'd break for users who either:
- Update from 0.7.x
- Update from 0.8.x **but** did not enable experimental features & update their compose with the warning prompt

In both of these cases the QMP port is not there even though we expect it to be there, we'll have to think about how to address it.

I'm closing these PR-s (you seem to have two of them doing the same thing) in favour of our branch.

PR #172, opened 2025-09-23
feat: Add noVNC as Preset App

Levev:
feat: Add a way to specify custom behavior for preset app
Added , which maps a preset app path to a callback, which gets called if found in Winboat.launchApp
This shouldn't be a long term solution though, as AppManager needs a refactor.


TibixDev:
LGTM, thank you! 🙌

Issue #146, opened 2025-09-18
[Bug] WinBoat creating / install - error 3389: bind: address already in use


larieu:
### What version of WinBoat are you running?

WinBoat Beta v0.7.12 Prod

### Your Environment

Fedora release 41 (Forty One)
Gnome 47  / Wayland
FreeRDP Client	com.freerdp.FreeRDP	3.17.1	stable	flathub	system

### Steps to Reproduce / Context

Started the AppImage 
installed/  configured all the requirements.
when all green
Tried to start a Win 11 Pro
( when failed tried 10 enterprise and other, too ) 

the error in the file install.log is

> 2025-09-18 09:14:31 | ERROR | Command failed: docker compose -f /home/larieu/.winboat/docker-compose.yml up -d
>  Volume "winboat_data"  Creating
>  Volume "winboat_data"  Created
>  Container WinBoat  Creating
>  Container WinBoat  Created
>  Container WinBoat  Starting
> Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint WinBoat (08ae6e3809c2e8f98f8cdff2a9dd480a653956480a73fedb7d22142c275ece8b): error while calling RootlessKit PortManager.AddPort(): listen tcp4 0.0.0.0:3389: bind: address already in use

Closed
Tried to change in the dockker-compose.yml file 
from

>     ports:
>       - "8006:8006"
>       - "7148:7148"
>       - "3389:3389/tcp"
>       - "3389:3389/udp"

to

>     ports:
>       - "8006:8006"
>       - "7148:7148"
>       - "3489:3389/tcp"
>       - "3489:3389/udp"

with no success

### Logs

2025-09-18 09:14:30 | INFO | Starting installation...
2025-09-18 09:14:30 | INFO | New state: "Creating Compose File"
2025-09-18 09:14:30 | INFO | Creating compose file at: /home/larieu/.winboat/docker-compose.yml
2025-09-18 09:14:30 | INFO | Compose file content: {
  "name": "winboat",
  "volumes": {
    "data": null
  },
  "services": {
    "windows": {
      "image": "ghcr.io/dockur/windows:4.35",
      "container_name": "WinBoat",
      "environment": {
        "VERSION": "10e",
        "RAM_SIZE": "8G",
        "CPU_CORES": "2",
        "DISK_SIZE": "64G",
        "USERNAME": "larieu",
        "PASSWORD": "tecknicuss",
        "HOME": "${HOME}",
[...]

PR #181, opened 2025-09-25
Handle used ports during/after the installation

Levev:
This PR aims to solve #146. During setup we don't check whether the ports used by the container are open, which leads to this issue.

After the installation has completed, we should still check the availability of used ports when we start up the container.
We could notify them with a dialog perhaps.


Levev:
All there's left to do is to make the QMP connection use our PortManager infrastructure and we'll be able to review!

TibixDev:
🤞This should help users who were having port conflicts.

Issue #71, opened 2025-09-08
nftables support


xxnuo:
Arch Linux, has migrated its default firewall backend from iptables to nftables.
New kernel module: The kernel module corresponding to nftables is nf_tables.

Backward compatibility: To allow programs depending on the old iptables commands (such as Docker) to continue working, the system provides a tool called iptables-nft. This tool offers the exact same command-line syntax as the old iptables, but it converts these rules into nftables rules and loads them into the kernel.

```bash
which iptables-nft
/usr/bin/iptables-nft
which iptables
/usr/bin/iptables
```

xxnuo:
#51

xxnuo:
lsmod | grep iptable_nat
failed

M0n7y5:
I am currently unable to use this on Arch based linux (CachyOS), is there any workaround for this?

Cris70:
I have the same problem on openSuse Tumbleweed, but I was able to make it work after loading the iptable_nat kernel module.
Since Tumbleweed does not make use of iptables (it uses nftables and the iptables-nft tool to trick programs which look for iptables), loading iptables_nat has no side effects, but it is enough for Winboat to happily pass the checks.

xxnuo:
> I am currently unable to use this on Arch based linux (CachyOS), is there any workaround for this?

I recompiled winboat and removed the iptable_nat check, and it now works properly.

ThomasCr:
I have the same problem with Linux Mint 22.1  / Ubuntu 24.04.3 (Noble Numbat)
Modern Linux System switches to nftables.

Nycz-lab:
same problem on nixos. this resolves it

PR #188, opened 2025-09-27
Check nf_tables when iptables check fails

naihe2010:
https://github.com/TibixDev/winboat/issues/71


TibixDev:
#210 is also related it seems, I'll look into it.

TibixDev:
Hey, this issue is not relevant anymore since with the version 5.07 of the `dockur/windows`, there's now a shortcut to the shared drive, and these kernel modules only help with network *discovery* and speed things up slightly, but they are not a hard requirement according to the creator of  `dockur/windows`. I removed the check entirely in fc8a006.

Thank you for contributing anyway and I hope you can find some other issue to solve. :)

Issue #171, opened 2025-09-23
[Bug] Cannot create container under NIX


fuma-afk:
### What version of WinBoat are you running?

v 0.8.3

### Your Environment

Linux distro used: Nixos unstable
DE: Hyprland
Version of xrdp: latests available

### Steps to Reproduce / Context

Steps to reproduce: after installing the flake i am able to launch winboat, get all green flags (after today's latest update that fixed docker user group), but when actually try to create the container the installation fails.

install log pasted below

### Logs
```
2025-09-23 17:49:31 | INFO | Starting installation...
2025-09-23 17:49:31 | INFO | New state: "Creating Compose File"
2025-09-23 17:49:31 | INFO | Creating compose file at: /home/fuma/.winboat/docker-compose.yml
2025-09-23 17:49:31 | INFO | Compose file content: {
  "name": "winboat",
  "volumes": {
    "data": null
  },
  "services": {
    "windows": {
      "image": "ghcr.io/dockur/windows:4.35",
      "container_name": "WinBoat",
      "environment": {
        "VERSION": "11",
        "RAM_SIZE": "8G",
        "CPU_CORES": "2",
        "DISK_SIZE": "40G",
        "USERNAME": "nope",
        "PASSWORD": "placeholder",
        "HOME": "${HOME}",
        "LANGUAGE": "Italian",
        "HOST_PORTS": "7149",
        "ARGUMENTS": "-cpu host,arch_capabilities=off \n-qmp tcp:0.0.0.0:7149,server,wait=off\n"
      },
      "cap_add": [
        "NET_ADMIN"
      ],
      "privileged": true,
      "ports": [
        "8006:8006",
        "7148:7148",
        "7149:7149",
        "3389:3389/tcp",
        "3389:3389/udp"
      ],
      "stop_grace_period": "120s",
      "restart": "on-failure",
      "volumes": [
        "data:/storage",
        "${HOME}:/shared",
        "/dev/bus/usb:/dev/bus/usb",
        "./oem:/oem"
      ],
      "devices": [
        "/dev/kvm"
      ]
    }
  }
}
2025-09-23 17:49:31 | INFO | New state: "Creating OEM Assets"
2025-09-23 17:49:31 | INFO | Creating OEM assets
2025-09-23 17:49:31 | INFO | Created OEM directory: /home/fuma/.winboat/oem
[...]

Issue #164, opened 2025-09-22
[Bug] Nix build isn't working


0xDracula:
### What version of WinBoat are you running?

last version

### Your Environment

Nixos
Kde Plasma
3.17.0

### Steps to Reproduce / Context

Try to install it on nix, after #55 

### Logs

..

### Expected Behavior

It should run without any problems

### Current Behavior

```
error: hash mismatch in fixed-output derivation '/nix/store/i46hmsnl0gps8723mj3k1ir8cdlp77r2-winboat-0.8.2-x86_64.AppImage.drv':
         specified: sha256-ViqXfoqyphllhQd40mL4B/x9l/p4065b4n0UmTGjFKc=
            got:    sha256-C0RJQFnpzV1RlggLSHSNb5j7AThJIr9grD3pcnCsBQQ=
error: 1 dependencies of derivation '/nix/store/7kglwp4lqmly32csqsr2jk7ringzqcng-winboat-0.8.2-extracted.drv' failed to build
error: 1 dependencies of derivation '/nix/store/8mc61rrjgpliwfw3y7q52r9ri0q9liz9-winboat-0.8.2.drv' failed to build
error: 1 dependencies of derivation '/nix/store/3bp33wglqw41wzsb6paqlkcbs7kd5lz8-system-path.drv' failed to build
error: 1 dependencies of derivation '/nix/store/hilb0fgj21cqbrw776z8dcp0642m8jg3-nixos-system-nixos-25.11.20250908.b599843.drv' failed to build
```

### Possible Solution

Change the fixed hash in the flake.nix

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

0xDracula:
I've forked it locally, and changed the SHA-256 and it's still complaining, I don't have any clue

TibixDev:
Can you check with v0.8.3? I triggered the workflow manually for now.

EliasBorrajo:
@0kenx Could maybe have a fix for that ? He is the one who hardcoded the sha256 value into the repo.

0xDracula:
> Can you check with v0.8.3? I triggered the workflow manually for now.


same issue, I've already forked it locally, and tried to change it myself, and same problem

realitymolder:
> > Can you check with v0.8.3? I triggered the workflow manually for now.
> 
> same issue, I've already forked it locally, and tried to change it myself, and same problem


I used my own script, you can try it / learn from it, it built for me:

```
[...]

PR #170, opened 2025-09-23
Add nix install instructions to README.md

realitymolder:
I've noticed the new flake.nix is missing installation instructions in the README so I've added a small section.

PR #194, opened 2025-09-28
Nix support 2.0

Simon-Weij:
I packaged winboat again, with a couple of changes.

Over the current flake:
Packages the tarball instead of the appImage, this solves a couple isolation issues (like not detecting the docker group)

Uses the tarball from the release, this will eventually make it easier to merge winboat into nixpkgs

Doesn't package freerdp, this should help with a bug where freerdp showed a black box and instantly crashes (see note)

It tries to keep stuff pretty minimal so there won't be many nix-specific bugs

Workflow to update the hash, which I've tested and should work fine. But obviously it hasn't been tested on the main repository

This pr should fix issue #171 and #164 

NOTE:
The black box still happens sometimes, this package works 100% fine on my desktop, but the bug occurred on my laptop, which later resolved itself, I'm not entirely sure how to reproduce this, but this version should work for more people than the current version that packages the appImage.

### DIFF
--- a/.github/workflows/update-flake-hash.yml
+++ b/.github/workflows/update-flake-hash.yml
@@ -58,12 +58,12 @@ jobs:
             git config user.email "github-actions[bot]@users.noreply.github.com"
             echo "set user.email to github-actions[bot]@users.noreply.github.com"
 
-            git fetch origin nix-support
-            git merge --no-edit -X theirs origin/nix-support
+            git fetch origin main
+            git merge --no-edit -X theirs origin/main
 
             sed -i "s/sha256 = \".*\";/sha256 = \"${ESC_HASH}\";/" flake.nix
 
 
             git add flake.nix || true
             git commit -m "chore: update flake.nix for version ${VERSION}" || echo "No changes to commit"
-            git push origin HEAD:nix-support || echo "Push failed"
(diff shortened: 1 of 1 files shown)


Simon-Weij:
Note about the note, the black box bug most likely has nothing to do with the nix package, but it was a thing of freerdp when you have the windows desktop open, it can't open apps.

TibixDev:
Thanks, hopefully this will work better for Nix folks. 🙏

xrexy:
@TibixDev PR didn't close the related issues

realitymolder:
@Simon-Weij thanks for the hard work. Seems like you removed the support of the module? 
Before the nix flake would allow the user to call ` services.winboat.enable = true;` and it took care of everything for the user. 
I'm asking because I want to write those things in the docs :)

Etmix:
#170 Handles the documentation.
We should probably bring back the module and validate that rootless is disabled to avoid the issue in #171.

Simon-Weij:
This flake is pretty much from scratch again, I or someone else can re-add the module if you ask me.

ppom0:
I was close to finishing packaging winboat from source for Nix, you may be interested: https://github.com/NixOS/nixpkgs/issues/440321
I think with current flake, my attempt, and last module, an upstream package can be made.

PR #209, opened 2025-10-02
Added desktop shortcuts

Simon-Weij:


Levev:
First of all, thank you for taking your time to implement this feature!
I see that you've decided to create a shell wrapper around the FreeRDP command which would work for some cases but there are some caveats to this approach.

You see, the WinBoat process manages a lot of things on top of the docker container and FreeRDP - one good example is the USB passthrough feature. The WinBoat process listens for usb connect/disconnect events, then interacts with the QEMU instance through a QMP socket. We could include this functionality in the shell wrapper, but as time goes on we'd basically have to implement WinBoat features twice, both in the app itself and the wrapper. (another example for something that wouldn't work with the wrapper approach is PR #181)

The implementation we'd aim for is to invoke WinBoat from the .desktop files directly, with a command line argument specifying the application to be ran. I won't go into the details here as this comment is already getting a bit too long, but this way we could use the already existing infrastructure to handle opening apps.

So far we didn't implement this feature due to the fact that our packaging solutions aren't quite consistent. For example, if someone uses the AppImage through say Gear Lever, then this solution would break down as winboat wouldn't be included in the path or .bashrc/.zshrc.

I won't be closing this PR just yet, I'll wait for @TibixDev to decide as to how to proceed.

PR #213, opened 2025-10-04
expose dockur only to 127.0.0.1

Zoey2936:


Leseratte10:
I haven't tested it yet, but this will probably break setups where the system is being accessed over IPv6, and even if it doesn't, restricting it to IPv4-only sounds like a bad idea in 2025. Can [::1] (IPv6 localhost) be added as well if this is implemented?

Zoey2936:
`[::1]` can be added the same way at the same time, but then it might not work if ipv6 is disabled. The question is if the ports are ever used over ipv6?
```
PC:~/winboat# grep -rne "::1"
PC:~/winboat# grep -rne "localhost"
src/main/main.ts:103:        mainWindow.loadURL(`http://localhost:${rendererPort}`);
PC:~/winboat# grep -rne "127.0.0.1"
src/renderer/lib/qmp.ts:116:     * @param host - The hostname of the qmp connection (e.g. 0.0.0.0, 127.0.0.1)
src/renderer/lib/winboat.ts:413:            this.qmpMgr = await QMPManager.createConnection("127.0.0.1", QMP_PORT).catch(e => {logger.error(e); throw e});
src/renderer/lib/winboat.ts:609:        /v:127.0.0.1\
src/renderer/lib/winboat.ts:628:                /v:127.0.0.1\
src/renderer/lib/constants.ts:8:export const WINBOAT_GUEST_API = "http://127.0.0.1:7148";
src/renderer/lib/install.ts:265:                const response = await nodeFetch('http://127.0.0.1:8006/msg.html');
src/renderer/views/SetupUI.vue:544:                            In the meantime, you can grab a coffee and check the installation status <a href="http://127.0.0.1:8006" @click="openAnchorLink">in your browser</a>.
src/renderer/App.vue:29:                            <a @click="openAnchorLink" href="http://127.0.0.1:8006/" target="_blank" rel="noopener noreferrer">http://127.0.0.1:8006/</a>
```

Leseratte10:
Well the "loadURL" from localhost could then start to fail on certain systems if Happy Eyeballs isn't working, since localhost usually resolves to ::1 on most systems. 

I know not everyone has IPv6 internet connectivity but are there still Desktop Linux systems out there where people hard-disable IPv6 in a way that not even a "ping ::1" works?

Zoey2936:
> I know not everyone has IPv6 internet connectivity but are there still Desktop Linux systems out there where people hard-disable IPv6 in a way that not even a "ping ::1" works?

localhost should fall back to 127.0.0.1 in case IPv6 is disabled and yes, sadly there are still people who disable IPv6 for some reason... But localhost could also be replaced with 127.0.0.1 in the source code

Levev:
Hey, thanks for taking your time to help out, but with the 0.9.0 release this has already been taken care of!

PR #220, opened 2025-10-05
chore: update flake.nix for version 0.8.6-alpha

xenicle:
Adapted SHA256 value for 0.8.6-alpha


TibixDev:
Hey, so basically 0.8.6 is an alpha for now, it will be updated manually once we release it. The stable release and the nix flake corresponds to 0.8.5.

Issue #228, opened 2025-10-07
[Feature] Class Prefix


alexankitty:
### Feature Description

Allow setting a class prefix for the /wm-class: argument of xfreerdp

### Use-Case

It would make it substantially easier to create window rules that specifically target xfreerdp windows spawned by winboat.

### Proposed Implementation

At the bare minum the class should be prefixed with winboat: followed by the name of the program or similar.
Potentially a configuration option could be set to allow changing the class prefix in the event a user needs a different class prefix for a conflict or whatever reason.

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [x] I would like to help out implementing this feature

waffles-dev:
Implemented in #229

PR #229, opened 2025-10-07
Implement wm prefix config and option

alexankitty:
Creates an implementation for setting the /wm-class: xfreerdp option, and a wmClassPrefix config to go alongside it in wbconfig.
Should implement #228 

Sets the default class prefix to `winboat-`

### DIFF
--- a/src/renderer/lib/config.ts
+++ b/src/renderer/lib/config.ts
@@ -11,6 +11,7 @@ export type WinboatConfigObj = {
     passedThroughDevices: PTSerializableDeviceInfo[];
     customApps: WinApp[]
     experimentalFeatures: boolean
+    wmClassPrefix: string
 };
 
 const defaultConfig: WinboatConfigObj = {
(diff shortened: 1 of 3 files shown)


TibixDev:
Hey, thanks for the contribution! Do you think there's any harm in having this transparently enabled in the background without the option?

alexankitty:
The only thing I could potentially see being an issue is if for some reason it conflicted with someone's rules for who knows why. It was just excessive forward thinking on my part. It's pretty unlikely, so really it's up to you.

TibixDev:
Yep I think we can have it enabled by default and no need for the option that way.

alexankitty:
I'll get it removed and rebased.

alexankitty:
I ended up just resetting my branch and then added the wm-class prefix. Let me know if that will work.

TibixDev:
Hey, sure that'll work yep, could you just resolve the merge conflict? I'll merge it right away after.

alexankitty:
Rebased. should be resolved now.

Issue #79, opened 2025-09-09
Apps only render on one monitor at a time


amphyvi:
Hi, apps only seem to render on one monitor at a time. If I move them to another monitor, they stop rendering and stop responding entirely. As long as the app is 100% within the monitor it opens on, it works 100% - but if (for example) 10% of the app's window ends up on another monitor, that 10% will be all white and entirely unresponsive.

Is there a fix for this? Or is this a limitation with Winboat/FreeRDP/Docker/etc?

Operating System: Ubuntu Studio 25.04
KDE Plasma Version: 6.3.4
KDE Frameworks Version: 6.12.0
Qt Version: 6.8.3
Kernel Version: 6.14.0-29-generic (64-bit)
Graphics Platform: Wayland
Processors: 24 × 13th Gen Intel® Core™ i7-13700K
Memory: 62.6 GiB of RAM
Graphics Processor: AMD Radeon Graphics
Manufacturer: ASUS

Levev:
I only have a single monitor sadly so I can't really test this out, but seems FreeRDP's `/multimonitor` flag could solve this issue.

amphyvi:
Sounds promising, thanks for the recommendation. How would I use that flag? I'm familiar with using flags with terminal applications but Winboat seems to be GUI only (or at least, primarily).

Ericchi:
I also have this issue, with the added bonus that if the two screen have different scaling factors, everything becomes extra-glitchy and the mouse doesn't work properly anymore since it can only reach some parts of the window.
The fact that windows spawn between the two screens adds even more issues to the whole experience.

Here's a screenshot of what happens if I drag the windows from a screen to the other:
<img width="1825" height="1094" alt="Image" src="https://github.com/user-attachments/assets/b2489940-5c28-48a4-9004-af5cbf75ae68" />

Levev:
> Sounds promising, thanks for the recommendation. How would I use that flag? I'm familiar with using flags with terminal applications but Winboat seems to be GUI only (or at least, primarily).

Oh, I've meant that we should look into adding this flag to the freerdp command that gets invoked upon opening an app.

VanillaCold:
Same issue here, on Fedora 42 with KDE Plasma 6.4.4; two monitors, different scaling on both of them. Apps appear on my secondary (1080p@75hz) monitor and can't be interacted when moving to my primary (1440p@180hz) screen.

TibixDev:
[...]

PR #230, opened 2025-10-07
Adds a multi monitor toggle

alexankitty:
Similar to my other pull request in implementation.
This adds an option that allows toggling the /multimon flag for xfreerdp3.
/multimon is required to have windows applications render on screens other than the main screen. It works, but moving applications between monitors forces it to redraw which makes it flash white for a moment.

### DIFF
--- a/src/renderer/views/Config.vue
+++ b/src/renderer/views/Config.vue
@@ -297,7 +297,7 @@
                         <div class="flex flex-row gap-2 items-center mb-2">
                             <Icon class="inline-flex text-violet-400 size-8" icon="uil:monitor"></Icon>
                             <h1 class="my-0 text-lg font-semibold">
-                                Multi Monitor Support
+                                Multi-Monitor Support
                             </h1>
                         </div>
                         <p class="text-neutral-400 text-[0.9rem] !pt-0 !mt-0">


TibixDev:
Hey, thanks for your contribution!

So I personally don't have multiple displays, so I was not able to test these fixes. There's a big issue #79 regarding this problem specifically, there we recommended the `+span` and `/multimon` flags but not everyone reported back positive results. Some people said `+span` works, others said `/multimon` works, some said it *kinda* works, others said neither works, so not exactly sure.

I don't wanna ship anything broken, even so I'm leaning towards merging this to help the users because it'd be better than not having any multi monitor support, but could I hear your feedback? Perhaps you could test with different resolutions and refresh rates on both displays for a wider scenario and figure out how to achieve the correct behaviour?

alexankitty:
Testing with both `/multimon` and `+span` seem to work well for me.
Below is my monitor setup
image
HDMI-A-1 runs at 1920x1080@60hz rotated 90°
DP-1 and DP-2 run at 2560x1440@144hz

The only issue I really run into is on my third display (both `+span` and `/multimon`) causes the app to throw itself to a different screen, however that is an issue with hyprland's display order under xorg and has nothing to do with xfreedrp3

If it would be preferred though, I could add a toggle for `+span`

alexankitty:
Something I should probably mention while it's on my mind. I had a whole host of problems initially that required some window rules to correct which is also why I put in #229 

Here are the rules I had to use for hyprland to get things stable
```
windowrule = suppressevent fullscreen maximize fullscreenoutput, class:^(winboat-.*)$
windowrule = nomaxsize, title:PopupHost
```

alexankitty:
I updated this PR to instead implement the multi monitor support as a drop down so that the end user may select which option works best for their setup. 
image

TibixDev:
Yeah this is nice I feel like it's the best of both worlds, thank you! 🙌

JSH32:
For some reason it doesn't open on my KDE Plasma desktop when using either, the buttons don't open up any applications
image

PR #238, opened 2025-10-08
Fixed typo in SetupUI

dxrkinfuser44:
changed "In the meantime you can grab coffee and check the status" to "In the meantime, you can grab a coffee and check the installation status" (line 515, ./src/renderer/views/SetupUI.vue)


TibixDev:
Thanks!

PR #260, opened 2025-10-09
Fix typo in home folder sharing acknowledgment message

jimjohnson-dev:
Corrected the spelling of "acknowledge" in the home folder sharing acknowledgment message.


TibixDev:
I think #290 already did this, but thanks either way :smile:

Issue #205, opened 2025-10-02
[Feature] mac os intel


dawnadvent:
### Feature Description

when the intel mac os can use it

### Use-Case

when the intel mac os can use it

### Proposed Implementation

when the intel mac os can use it

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [x] I would like to help out implementing this feature

kaystrobach:
docker for mac can run x86 images - so having this on macos should not be a problem at all. just the electron app needs to be cross compiled ... thanks

bojogon7:
Link to docker MacOS x86 project

https://github.com/dockur/macos

PR #274, opened 2025-10-09
First step to add Mac OS Support 

ahmedbarakat207:


Levev:
Converted to draft, please mark as ready for review when you're done.

Thank you for your work on getting WinBoat to work under MacOS!

src/renderer/lib/specs.ts
```diff
@@ -95,8 +108,12 @@ export async function getSpecs() {
 
     // Docker user group check
     try {
-        const userGroups = (await execAsync('id -Gn')).stdout;
-        specs.dockerIsInUserGroups = userGroups.split(/\s+/).includes('docker');
+        if (os.platform() === 'darwin') {
+            specs.dockerIsInUserGroups = true;
```
Hi-Angel, line 112: Is that guaranteed on Mac? And I think it warrants a comment.

src/renderer/lib/specs.ts
```diff
@@ -32,23 +32,31 @@ export async function getSpecs() {
...
+        }
     } catch (e) {
         console.error('Error reading /proc/meminfo:', e);
     }
 
     // KVM check
     try {
+        if (os.platform() === 'darwin') {
+            specs.kvmEnabled = true;
+        }
```
Hi-Angel, line 59: I don't think it's correct. Looking [at this post](https://stackoverflow.com/q/46565353/2388257), access to `/dev/kvm` may or may not exist depending on user configuration, so the check for file existence a few lines below is necessary. You can't just assume the file is accessible on 'darwin'.

Hi-Angel (changes requested):
Please add to the commit a:

Fixes: https://github.com/TibixDev/winboat/issues/205

And also the second commit is weird, it fixes the first one and it adds a memory change… and what "add files via upload" commit name would even mean?

Issue #177, opened 2025-09-24
[Bug] Workflow "update-flake-hash" not triggering on releases


xrexy:
### What version of WinBoat are you running?

0.8.4

### Steps to Reproduce / Context

Trigger a github Release

### Logs

https://github.com/TibixDev/winboat/actions/workflows/update-flake-hash.yml

### Expected Behavior

On every release the workflow should trigger and update the appimage hash in `flake.nix`. 

Ran the commands locally and everything is working as expected, seems the issue is only the triggering.

### Current Behavior

On a release the workflow is not triggering

### Possible Solution

Most likely related to:
```
on:
  release:
    types: [published]
  workflow_dispatch:
    inputs:
      version:
        description: 'Version to update (e.g., 0.8.1)'
        required: true
```

Related to #164 

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

TibixDev:
I can confirm this is an issue, I had to trigger it manually for the last two releases. 🙁

0kenx:
Yeah I've not tested the triggering part of the action.... I only tested a simplified local action for the hash calculation.

xrexy:
Found this in their documentation:

> When you use the REST API or GitHub CLI to create a release, the release event is triggered. However, if a release is created by a workflow run, no new workflow run is triggered by the release event to prevent infinite recursion.

So, if this is still the case(which it looks like it is) I believe only fix would be to merge this workflow inside the release one

Levev:
I'll be closing this as not planned since we got upstreamed with NixOS/nixpkgs#451068

PR #277, opened 2025-10-09
fix: update release tags with corrected flake hash

### LINKED ISSUE #164: [Bug] Nix build isn't working

jordangarrison:
## Problem

Users experience hash mismatch errors when trying to run WinBoat from release tags:

```bash
nix run github:TibixDev/winboat/v0.8.7
```

Error:
```
error: hash mismatch in fixed-output derivation '/nix/store/...-winboat-0.8.7-x64.tar.gz.drv':
         specified: sha256-I7SoX0dcQoFucj/GRvORZvgdKTIc+7+F9COZS33nFuk=
            got:    sha256-4NV9nyFLYJt9tz3ikDTb1oSpJGAKr1I49D0VHqpty3I=
```

## Root Cause

The `update-flake-hash` workflow has a timing issue:

1. Release tag is created with an **outdated hash** from `flake.nix`
2. Release artifacts are built and published
3. Workflow calculates the correct hash and updates **main branch**
4. **Tag is never updated**, so it still contains the wrong hash

When users reference a tag (e.g., `v0.8.7`), Nix uses the `flake.nix` from that tag, which has the incorrect hash.

### Evidence
- Tag `v0.8.7` (commit `a641ec8`) has hash: `1s8nwxylp693yj2vzyqw68livy36j7rldiizf9p82hjw8xgsid13` (wrong - this is for v0.8.6)
- Main branch has hash: `0wnbdnm1w59xyhw55bqac0jak16nvcs91qixnxyrnq2b46gpvmg0` (correct for v0.8.7)

## Solution

This PR modifies `.github/workflows/update-flake-hash.yml` to force-update the release tag after calculating and committing the correct hash. This ensures the tag points to a commit with the accurate hash matching the release artifacts.

## Changes

- Added logic to force-update the release tag with the corrected `flake.nix`
- Tag now points to the commit with the matching hash for the release artifacts

## Testing

Locally tested:
- ✅ Hash calculation matches the published v0.8.7 artifacts
- ✅ Sed replacement logic works correctly
- ✅ Tag force-update logic verified (creates and moves tags successfully)
- ✅ Built successfully from local clone with correct hash

## Impact

- Fixes hash mismatch errors for all future releases
- Users can reliably use `nix run github:TibixDev/winboat/vX.Y.Z`
- Existing tags (v0.8.6, v0.8.7) will need manual correction or the workflow can be triggered manually to fix them

Fixes #164

### DIFF
--- a/.github/workflows/update-flake-hash.yml
+++ b/.github/workflows/update-flake-hash.yml
@@ -1,8 +1,12 @@
 name: Update Flake Hash on Release
 
 on:
-  release:
-    types: [published]
+  workflow_run:
+    workflows: ["Build WinBoat"]
+    types:
+      - completed
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       version:
@@ -12,6 +16,10 @@ on:
 jobs:
   update-hash:
     runs-on: ubuntu-latest
+    # Only run if the workflow succeeded and was triggered by a tag
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      startsWith(github.event.workflow_run.head_branch, 'refs/tags/')
     permissions:
       contents: write
       
@@ -35,16 +43,33 @@ jobs:
           script: |
             set -euo pipefail
 
-            if [ "${{ github.event_name }}" = "release" ]; then
-              VERSION="${{ github.event.release.tag_name }}"
+            if [ "${{ github.event_name }}" = "workflow_run" ]; then
+              # Extract version from the tag ref (refs/tags/v0.8.7 -> v0.8.7)
+              VERSION="${{ github.event.workflow_run.head_branch }}"
+              VERSION="${VERSION#refs/tags/}"
             else
-              VERSION="${{ github.event.inputs.version }}"
+              # Manual workflow_dispatch trigger
+              VERSION="${{ inputs.version }}"
             fi
 
             VERSION="${VERSION#v}"
 
             URL="https://github.com/TibixDev/winboat/releases/download/v${VERSION}/winboat-${VERSION}-x64.tar.gz"
 
+            echo "Checking if release asset is available..."
+            for i in {1..30}; do
+              if curl --head --fail "$URL" &>/dev/null; then
+                echo "✓ Release asset is available"
+                break
+              fi
+              if [ $i -eq 30 ]; then
+                echo "✗ Release asset not available after 5 minutes"
+                exit 1
+              fi
+              echo "Waiting for asset... (attempt $i/30)"
+              sleep 10
+            done
+            
             echo "Prefetching from: $URL"
             HASH=$(nix-prefetch-url "$URL")
 


jordangarrison:
Note: This PR also relates to #177, which reports that the `update-flake-hash` workflow isn't triggering on releases.

While this PR fixes the tag update logic (ensuring tags point to commits with correct hashes), there may be a separate issue with the workflow trigger itself. If the workflow isn't running at all, then this fix won't help until that's resolved.

The workflow trigger is currently:
```yaml
on:
  release:
    types: [published]
```

If the workflow isn't triggering, we may also need to investigate:
1. Whether releases are being published (vs created as drafts)
2. Repository permissions for GitHub Actions
3. Whether the workflow file exists on the default branch when releases are created

Both issues should be addressed for a complete fix.

jordangarrison:
## Update: Now also fixes #177

I've updated this PR to comprehensively address both issues:

### Root Cause Analysis

The workflow wasn't triggering automatically because:
1. Releases created by `GITHUB_TOKEN` don't trigger `release: published` events (GitHub Actions limitation to prevent infinite loops)
2. Tags had incorrect hashes because the workflow only updated main, never the tags

### Complete Solution

**Changed trigger mechanism:**
- ✅ Replaced `release: published` with `workflow_run` 
- ✅ Triggers after "Build WinBoat" workflow completes
- ✅ Only runs on successful workflows triggered by tags
- ✅ Extracts version from `workflow_run.head_branch`

**Improved reliability:**
- ✅ Replaces `sleep` with proper retry loop using `curl` to verify asset availability
- ✅ Polls up to 30 times (5 minutes) with clear error messages
- ✅ Keeps `workflow_dispatch` for manual triggers

**Tag update (original fix):**
- ✅ Force-updates release tags to point to commits with correct hashes

### Testing

All workflow logic has been validated:
- ✓ Version extraction from tag refs
- ✓ URL format correctness  
- ✓ Retry loop logic
- ✓ Workflow trigger conditions
- ✓ Asset availability checking

### Impact

This fix ensures:
- Workflow runs automatically after each tagged release
- No more manual workflow_dispatch triggers needed
- Release tags always have correct flake hashes
- `nix run github:TibixDev/winboat/vX.Y.Z` works reliably

Fixes #164 and #177

Simon-Weij:
Looks good to me

waffles-dev:
Looking at this one (Simon tells me it's still needed, I'm sorry for the delay) - can you please catch up to latest/resolve conflicts and I'll add it to the queue to be looked at for merging - thanks!

waffles-dev:
flake removed in #434

PR #283, opened 2025-10-09
feat: add app filtering

devin-slothower:
**Small modification to allow filtering internal and custom applications.**

Screenshot_20251009_195507
Screenshot_20251009_195726

### DIFF
--- a/src/renderer/views/Apps.vue
+++ b/src/renderer/views/Apps.vue
@@ -115,7 +115,7 @@
                     :disabled="!winboat.isOnline.value"
                     class="flex flex-row-reverse gap-1 items-center justify-center"
                 >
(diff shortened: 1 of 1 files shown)


TibixDev:
Hello, this is a cool change! Just a couple small things, could you assign a different icon, something more akin to a filter? You can check which icons are supported this way on Xel's docs site or use Iconify if none of those icons are satisfying.

devin-slothower:
Spent a while struggling with iconify not playing nicely with Xel, not a big css guy, who knew you could reverse the flexbox direction :laughing: 

Anyways, got a nice filter icon now.

Screenshot_20251011_193517

TibixDev:
Yeah I also had Xel not play very well with Iconify before 😄
Thanks, it looks great!

PR #298, opened 2025-10-11
[QoL] Add resource allocation warnings and enhanced system detection

doughty247:
## Description

This PR adds quality-of-life improvements to the VM resource configuration experience by providing real-time feedback when users allocate system resources that may impact host performance.

## Motivation

Currently, WinBoat validates resource allocation with basic minimum checks (4GB RAM, 2 CPU cores) but doesn't warn users when they're allocating a large percentage of their available system resources. Users can inadvertently allocate too much RAM or CPU without realizing the impact on their host system until performance degradation occurs.

### 1. Yellow Warning System
- Added percentage-based warnings for RAM and CPU allocation in the Configuration view
- **RAM Warning**: Displays when allocation exceeds 40% of total system RAM
- **CPU Warning**: Displays when allocation exceeds 50% of total physical CPU cores
- Warnings appear as yellow text (`text-yellow-400`) directly below the respective input fields
- Example: `"⚠️ Using 65% of total RAM - high allocation may impact host performance"`

### 2. Enhanced System Detection
- Leverages existing `getMemoryInfo()` function in `specs.ts` that reads `/proc/meminfo` to extract both total and available system memory
- Returns `MemoryInfo` type with `totalGB` and `availableGB` properties
- Used by `getSpecs()` to provide accurate memory detection
- Enables real-time validation during both setup and configuration

### 3. Improved UI Components
- Replaced `<x-input>` components with native `<input type="number">` elements for RAM/CPU allocation
- Added custom styling to match WinBoat's design language:
  ```css
  class="w-24 text-right text-[1.1rem] bg-neutral-800 border border-neutral-600 rounded px-3 py-1.5"
  ```
- Changed card layout from `flex-row` to `flex-col` to accommodate inline warning messages

### File: `src/renderer/views/Config.vue`
- Added computed properties `ramWarning` and `cpuWarning`
- Calculate percentage of system resources being allocated
- Return warning strings when thresholds exceeded, otherwise return `null`
- Warning messages displayed conditionally with `v-if="ramWarning"` / `v-if="cpuWarning"`

### File: `src/renderer/lib/specs.ts`
- Already implements `getMemoryInfo()` function
- Reads `/proc/meminfo`, extracts `MemTotal` and `MemAvailable` lines
- Converts kilobytes to gigabytes with rounding: `Math.round(parseInt(...) / 1024 / 1024 * 100) / 100`
- Exports `MemoryInfo` type for type safety

## Benefits

1. **Proactive User Guidance**: Users receive immediate feedback about allocation impact before applying changes
2. **Prevents System Instability**: Warnings help users avoid over-allocating resources that could freeze their host system
3. **Better UX**: Native number inputs provide more familiar controls compared to custom components
4. **Consistent Validation**: Same detection logic used during initial setup and ongoing configuration

## Testing Notes

- Tested on systems with varying RAM amounts (16GB, 32GB)
- Warning thresholds trigger correctly at 40% RAM and 50% CPU
- Native number inputs respect min/max boundaries
- Warning messages display inline without breaking card layout
- Color scheme (`text-yellow-400`) provides appropriate visual hierarchy

## Screenshots

Screenshot_20251010_180636

---

## Notes

- All changes are backwards-compatible with existing WinBoat installations
- Warning system is non-blocking - users can still save their configuration choices

---

## Modified Files

- `src/renderer/views/Config.vue` (71 additions, 42 deletions)

**Total Changes**: ~113 lines modified

---

## Build Verification

✅ **Test Build Completed Successfully**
- Built clean AppImage
- No build errors or warnings
- Branch: `qol-warnings-and-detection`

### DIFF
--- a/src/renderer/views/Config.vue
+++ b/src/renderer/views/Config.vue
@@ -18,17 +18,16 @@
                                 How many gigabytes of RAM are allocated to the Windows virtual machine
                             </p>
                         </div>
-                        <div class="flex flex-row gap-3 justify-center items-center">
-                            <input
-                                type="number"
-                                class="w-24 text-right text-[1.1rem] bg-neutral-800 border border-neutral-600 rounded px-3 py-1.5 text-neutral-100 [&::-webkit-inner-spin-button]:ml-2 [&::-webkit-outer-spin-button]:ml-2"
+                        <div class="flex flex-row gap-2 justify-center items-center">
+                            <x-input
+                                class="max-w-16 text-right text-[1.1rem]"
                                 min="4"
                                 :max="maxRamGB"
-                                v-model.number="ramGB"
-                                step="1"
+                                :value="ramGB"
+                                @input="(e: any) => ramGB = Number(/^\d+$/.exec(e.target.value)![0] || 4)"
                                 required
-                            />
-                            <p class="text-neutral-100 font-medium">GB</p>
+                            ></x-input>
+                            <p class="text-neutral-100">GB</p>
                         </div>
                     </div>
                     <p v-if="ramWarning" class="text-yellow-400 text-xs mt-2 mb-0">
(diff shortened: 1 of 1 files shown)


Levev:
Thank you for taking your time to improve WinBoat's UX! I really like how the warning looks!

That being said, I'm not sure how I feel about removing x-card. Relying on xel-toolkit's components could seem like a limiting factor when designing the UI but in my opinion it is what makes WinBoat's look consistent. 
Also make sure you parse ramGB correctly, iirc we used a bit of RegEx for that.

doughty247:
Yeah... My idea there was that I wanted easy increase/decrease for users, but I do totally agree that it breaks the UI look. Thanks for the feedback! I will make the changes immediately.

doughty247:
Updated! I've switched back to x-input components with RegEx parsing as requested.

Changes made:
- Replaced native input elements with x-input components
- Added RegEx parsing: @input="(e: any) => ramGB = Number(/^\d+$/.exec(e.target.value)![0] || 4)"
- Kept the flex-col layout and all warning functionality
- Tested locally - everything works great!

Thanks for the feedback! @Levev

TibixDev:
This PR is neat idea, I have two little ideas/nitpicks in addition:
- When you install WinBoat, there's a warning for RAM allocation when you're trying to allocate more than available, could you also add that? The container will fail to start if that threshold is exceeded, the user technically has enough physical RAM for it but not enough available at that time. We should probably warn the user. you can take a look at SetupUI to see how this is implemented
- The threshold should probably be >60% RAM and >75% (3/4) of the cores

Issue #296, opened 2025-10-10
[Feature] Don't require .iso to launch container after install


Matheos96:
### Feature Description

May be related to #272 

I am using a custom iso, tiny11. I just noticed that after deleting the iso from disk, my container would no longer launch through winboat. That may make sense, but to me, as a dumb user, I would expect the .iso to no longer be needed once setup is completed.

### Use-Case

I don't want to keep a tiny11.iso forever in my downloads directory (where I unknowningly kept it during installation). Please let the container launch without it, if possible. If not possible, please explain why (please, I am not as dumb as I may sound :)) and let me relocate the .iso to another folder without reinstall.

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [ ] I would like to help out implementing this feature

heuzef:
Hi, same situation here.

Even after an unmount, the ISO file (7G) is stuck in my download folder :(

Any solution ?

auri-f5bde6:
For people who are having this issue, for now, as a workaround, go to `~/.winboat/docker-compose.yml`.
In the `    volumes:` section, delete the entry that have `.iso` mentioned in it.
Run `docker rm WinBoat`

You can either: 
1. Go though the winboat installation again (and use the same password, userlogin, but don't add iso) 
Or
2. Run `docker compose -f ~/.winboat/docker-compose.yml up -d`
Then try to launch it in winboat again!

*tell me if this work, as Im not 100% sure, thanks!

waffles-dev:
should be fixed in #315 please let me know if otherwise

Issue #272, opened 2025-10-09
[Feature] Detect when docker fails to start container due to missing iso file


GoldenReverie:
### Feature Description

I noticed when i removed the windows ISO from the folder my winboat instance failed to spin up in docker. it may not be obvious to most people that you might need to check docker for the reason why so id suggest some kind of error handling to catch when and why the docker instance will not spin up the container. Thanks

### Use-Case

Users with little docker experence.

### Proposed Implementation

Reading error messages when the docker command is sent and displaying this to the user via the GUI.

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [ ] I would like to help out implementing this feature

waffles-dev:
should be fixed in #315 please let me know if otherwise

PR #315, opened 2025-10-12
fix: unmount iso after install

AdianKozlica:
Addresses issues https://github.com/TibixDev/winboat/issues/296 and https://github.com/TibixDev/winboat/issues/272

### DIFF
--- a/src/renderer/lib/install.ts
+++ b/src/renderer/lib/install.ts
@@ -314,7 +314,7 @@ export class InstallManager {
 
                     if (config.services.windows.volumes.length !== filteredVolumes.length) {
                         config.services.windows.volumes = filteredVolumes;
-                        await winboat.replaceCompose(config)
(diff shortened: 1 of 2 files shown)


TibixDev:
Hey, this is a good change, but would it be possible to do this in a way where the install isn't marked as done until the restart is completed and we have the guest server online again? It could maybe be an extra step where the status says "Unmounting ISO" (only if you're installing with an ISO) and then back to waiting for Guest Server? Imo it would be better than saying it is finished and the user getting confused about why the Guest Server stays offline for 20-30s.

AdianKozlica:
Hello, the issue regarding the Guest Server being offline shouldn't occur since the container won't be restarted after the compose file has been edited.

TibixDev:
Ah you're right my bad, this is why you added the restart param. LGTM! 🙌

Issue #322, opened 2025-10-13
[Feature] Multiple instances of the same app


helpnoob:
### Feature Description

Being able to open multiple instances of the same windows application would be very useful for me.

### Use-Case

Some applications don't support internal multitasking so being able to have two windows open would be very helpful.

### Proposed Implementation

Shift-clicking the application in the WinBoat GUI would open another instance of that application.

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [ ] I would like to help out implementing this feature

PR #326, opened 2025-10-13
fix: run multiple apps

AdianKozlica:
Fixes https://github.com/TibixDev/winboat/issues/322

### DIFF
--- a/rdp_exec/main.go
+++ b/rdp_exec/main.go
@@ -3,7 +3,9 @@ package main
 import (
 	"encoding/base64"
 	"flag"
+	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


waffles-dev:
@AdianKozlica I think this might be a potentially exciting addition.

Can you please add more detail to the PR description? Make sure to list:
- What the problem was
- If this was a bug or a missing feature
- How you addressed it, include design decisions made - justify adding additional services/programs to the WinBoat guest vs extending existing
- Include any screenshots/recordings demonstrating the fix

Additionally please add comments/documentation in the new (go) code explaining briefly what it does

AdianKozlica:
Hello, so basically xfreerdp replaces the old instance of an app with a new one if the cli arguments are the same.

We can bypass this if we use a wrapper executable that has a dummy argument (in our case we use a random UUID).

Below is an example without the wrapper executable:

https://github.com/user-attachments/assets/2870a317-d9c9-4241-8264-5cebb10f2013

And below is the example with the wrapper:

https://github.com/user-attachments/assets/325cd5dc-1687-48ad-b263-2cfa3a57bfd3

src/renderer/lib/winboat.ts
```diff
@@ -677,7 +678,7 @@ export class Winboat {
         /scale-desktop:${this.#wbConfig?.config.scaleDesktop ?? 100}\
         ${combinedArgs}\
         /wm-class:"winboat-${cleanAppName}"\
-        /app:program:"${app.Path}",name:"${cleanAppName}" &`;
+        /app:program:"C:\\OEM\\rdp_exec\\rdp_exec.exe",cmd:"-cmd \\\"${app.Path}\\\" -dummy ${crypto.randomUUID().toString()}",name:"${cleanAppName}" &`;
```
waffles-dev: The `\` before each of the `"` chars looks unnecessary, can they be removed?

src/renderer/lib/winboat.ts
waffles-dev, line 1: when you catch up to current main, please run `npx prettier . --write` to ensure consistent formatting

src/renderer/lib/winboat.ts
AdianKozlica: Hello, the additional escapes are necessary in order to properly pass the command string, otherwise the freerdp command will fail

AdianKozlica:
Hello, I have made it possible so that if the user hasn't pressed the shift key, it will use "1" as the dummy argument, if the user has pressed the shift key it will generate a random UUID and it won't replace the current instance.

Note: explorer.exe will still be launched seperately despite it having the same dummy argument, this behaviour also happens when rdp_exec is not used.

waffles-dev:
Thankyou, is there anything special I need to do to test this? I dropped the new exe into the winboat dir in `C:\Program Files`, but they're still not launching.

Also, does this have any impact on users updating, or should they just need to update the guest using the zip?

I'll also note I get MS smartscreen blocking the app from running if I execute it in VNC - until I bypass it once. Could it be something like this preventing it running via RDP?

PR #332, opened 2025-10-14
feat: Add config options to add new or replace existing arguments for RDP

waffles-dev:
Added new Advanced Features toggle
image

Added new args and replacement args interface
image

New will just add a new argument to the RDP argument list
Replace will replace the argument matching the old_arg with the new_arg.

Note: Args are added/removed from config on input (as they're typed), if this is deemed too much, we can limit to on enter?


TibixDev:
Thanks for the PR, I believe it is in a great state now as we managed to iron out the issues on Discord, so merging this! 🙌

PR #336, opened 2025-10-14
Adds feature to support winboat's system tray app

ADIOR-enigma:
This PR aims to add feature support for winboat's system tray app, enabling persistent background operation and user-friendly interaction without keeping the main window open.
Uses Electron’s Tray and Menu modules along with nativeImage for seamless integration.
image


TibixDev:
Hey, this is a good feature but there's tons of issues that need to be addressed before something like this can be implemented:
- We have zero guarantees that Docker is even installed at the point of minimizing, because minimizing to tray is something you can do any time, which includes before or during the install phase
- We have zero guarantees for any of the implemented behaviour as outlined in the previous point
- Most apps with tray icons implement minimizing differently - the close button minimizes to tray (from where you can subsequently quit the app via the context menu) while the minimize button minimizes to taskbar **or** they ask you which one you prefer
- You're hardcoding the VNC UI port, in future versions we will have no guarantee for static ports (we will scan the compose file to find the host port)
- Much of the Docker specific code is duplicate of what the WinBoat (`winboat.ts`) class can do does - except we cannot rely on that class since again you can minimize to tray before/during install where even spawning that class is something that should not be done

Overall this PR has a solid concept but there's many-many considerations that need to be made and issues resolved so I can't merge this with confidence in this state.

adior-enigmac:
thanks for pointing out the issues, will keep in mind the considerations in future PRs, should i close this PR?

adior-enigmac:
did the changes needed

ADIOR-enigma:
> Hey, this is a good feature but there's tons of issues that need to be addressed before something like this can be implemented:
> 
>     * We have zero guarantees that Docker is even installed at the point of minimizing, because minimizing to tray is something you can do any time, which includes before or during the install phase
> 
>     * We have zero guarantees for any of the implemented behaviour as outlined in the previous point
> 
>     * Most apps with tray icons implement minimizing differently - the close button minimizes to tray (from where you can subsequently quit the app via the context menu) while the minimize button minimizes to taskbar **or** they ask you which one you prefer
> 
>     * You're hardcoding the VNC UI port, in future versions we will have no guarantee for static ports (we will scan the compose file to find the host port)
> 
>     * Much of the Docker specific code is duplicate of what the WinBoat (`winboat.ts`) class can do does - except we cannot rely on that class since again you can minimize to tray before/during install where even spawning that class is something that should not be done
> 
> 
> Overall this PR has a solid concept but there's many-many considerations that need to be made and issues resolved so I can't merge this with confidence in this state.

give a look now

PR #337, opened 2025-10-14
Desktop icon support

Simon-Weij:
I made this draft pull request with the intention of eventually adding desktop icon support

How I currently plan on doing this (might change)

Make winboat support launching apps using a CLI app (winboat launch notepad for example). It defaults to looking in the user's path, otherwise a prompt will show up that asks the user where winboat is installed. I also plan on adding a loading spinner while the app launches, since the VM might take a bit to boot.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "winboat",
-    "version": "0.8.5",
+    "version": "0.8.7",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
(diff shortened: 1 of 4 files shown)


MDefender:
es, it would be great if you could not only launch apps from the apps menu themselves, but also have the option to create direct launch icons on the Linux desktop or pin them to the taskbar. I only use a few programs on Windows  and I would be happy if I could launch them directly from the desktop.

PR #344, opened 2025-10-15
One-liner to easily build from source (cURL or wget) and install.sh

adeel26in:

I have made a file named install.sh that simplifies the process of building from source. Instead of doing the building yourself you just have to cURL or wget install.sh and build from source will be done for you. This highly improves user convenience and also simplifies development from source. It uses the branch "main" which ensures that it will be stable. (Please ignore my commits on October 14th 2025 those just pull an AppImage doesn't build from source and also I have deleted the pull request I did yeserday regarding this). NOTE: (Not tested yet)

Kind Regards,
adeel26in

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # ==================================================
-# 🚀 WinBoat Source Installer v1.8
+# 🚀 WinBoat Source Installer v1.2
 # Description: Builds and installs WinBoat from source.
 # Supported OS:
 #   Ubuntu, Debian, Pop!_OS, Kubuntu, Lubuntu, Ubuntu Budgie,


Levev:
Thank you for working on this feature!
I do think this could have it's place in the project, but I think we should support other distros(package managers) with this script, namely Arch based ones.

adeel26in:
Will this be added as a feature right now or do we not focus on this?

install.sh
```diff
@@ -0,0 +1,186 @@
...
+}
+DISTRO=$(detect_distro)
+echo -e "${YELLOW}🧠 Detected distro:${NC} $DISTRO"
+
+# --- Package installer function ---
+install_pkg() {
+    local pkg=$1
+    if ! command -v "$pkg" &>/dev/null; then
+        echo -e "${YELLOW}📦 Installing ${pkg}...${NC}"
+        case "$DISTRO" in
```
Levev, line 104: What about NixOS, Arch based distros, etc?

adeel26in:
I highly extended distro support and also for the distros you asked to add support for. Please review this again. (This is now v1.2)

Kind Regards,
adeel26in

waffles-dev:
> I highly extended distro support and also for the distros you asked to add support for. Please review this again. (This is now v1.2)
> 
> Kind Regards, adeel26in

I'm sorry, this doesn't make sense/work.
You basically have "if bazzite, it's fedora, therefore install via dnf", this won't work. It'd need to be brew install in this case.

But this sort of issue is kind of proving the point. You can't have a one-size fits all distros for something like this.

Sorry, this won't work on even the current contributors' machines -, and this type of trying to cover all in a single case doesn't scale. You also don't specify versions for some dependencies, and have hardcoded versions for others.

I'm not sure if this is just an AI generated script, or if it's adapted from something you've used before - but it's clearly untested for the distros it supposedly does support, and doesn't support many others.

We'd much prefer efforts for supporting simplifying building from source be directed to defining a dev container so that all contributors can immediately build and contribute with an identical build environment, without polluting their machines and without potentially introducing side effects to their build that would require efforts from other contributors to help diagnose.

I don't wish to turn away a potential contributor and I hope more generally you don't feel discouraged from this. We'll probably always feel understaffed. But we'd like contributions to fit within our overall goals for the direction WinBoat. I am happy to assist you if you have any questions.

PR #346, opened 2025-10-15
Add desktop shortcuts and related changes. 

Bl4ckk98:
They will check if winboat VM is online and start it if not. They also sync with custom rdp Args and settings.
On creation, icons are being polled.
Winboat is being bypassed in order to be able to open apps without launching the actual winboat UI.

**I need to point out that jq is a requirement for this.**

### DIFF
--- a/src/renderer/lib/winboat.ts
+++ b/src/renderer/lib/winboat.ts
@@ -703,62 +703,26 @@ export class Winboat {
 
     // New function by bl4ckk
     async createDesktopShortcut(app: WinApp): Promise<{ scriptPath: string; desktopPath: string; iconPath: string }> {
-        const fs: typeof import('fs') = require('fs');
-        const path: typeof import('path') = require('path');
         const os: typeof import('os') = require('os');
-        const guestApiUrl = `http://127.0.0.1:${GUEST_API_PORT}`;
-
-        const cleanAppName = app.Name.replace(/[,.'"]/g, "");
-        const appFile = app.Name.replace(/\s+/g, '_');
-
-        // Paths on best effort
-        const homeDir = os.homedir();
-        const binDir = path.join(homeDir, '.local', 'bin');
-        const appDir = path.join(homeDir, '.local', 'share', 'applications');
-        const iconDir = path.join(homeDir, '.local', 'share', 'icons', 'winboat');
-
-        fs.mkdirSync(binDir, { recursive: true });
-        fs.mkdirSync(appDir, { recursive: true });
-        fs.mkdirSync(iconDir, { recursive: true });
-
-        const scriptPath = path.join(binDir, `${appFile}.sh`);
-        const desktopPath = path.join(appDir, `${appFile}.desktop`);
-        const iconPath = path.join(iconDir, `${appFile}.png`);
-
-        // Extract icon
-        let iconValue = 'utilities-terminal';
-        if (app.Path && app.Path !== InternalApps.WINDOWS_DESKTOP && app.Path !== CustomAppCommands.NOVNC_COMMAND) {
-            try {
-                const formData = new FormData();
-                formData.append('path', app.Path);
-                const iconRes = await nodeFetch(`${guestApiUrl}/get-icon`, {
-                    method: 'POST',
-                    body: formData as any
-                });
-                if (iconRes.ok) {
-                    const iconB64 = await iconRes.text();
-                    fs.writeFileSync(iconPath, Buffer.from(iconB64, 'base64'));
-                    iconValue = iconPath;
-                }
-            } catch (e) {
-                logger.warn(`Failed to fetch icon for ${app.Name}: ${e}`);
-            }
-        }
-
-        // Get credentials and port
         const { username, password } = this.getCredentials();
         const freeRDPBin = await getFreeRDP();
 
-        // get stockArgs as string
-        const stockArgsStr = stockArgs.join(' ');
+        // Path for centralized script
+        const centralScriptPath = path.join(os.homedir(), '.local', 'bin', 'winboat-helper.sh');
 
-        // Generate launcher script to dinamically read configs
-        const launcherScript = `#!/usr/bin/env bash
+        // Creat script only if it does not exist
+        if (!fs.existsSync(centralScriptPath)) {
+            const centralScript = `#!/usr/bin/env bash
     set -euo pipefail
+
+    # WinBoat Centralized App Launcher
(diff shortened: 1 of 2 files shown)


electron-builder.json
Bl4ckk98, line 1: Just revert this one, it was making the tests faster

src/main/main.ts
```diff
@@ -56,14 +56,21 @@ let mainWindow: BrowserWindow | null = null;
...
-        // @ts-ignore property "window" is optional, see: [dialog.showMessageBoxSync](https://www.electronjs.org/docs/latest/api/dialog#dialogshowmessageboxsyncwindow-options)
-        dialog.showMessageBoxSync(null, {
-            type: "error",
-            buttons: ["Close"],
-            title: "WinBoat",
-            message: "An instance of WinBoat is already running.\n\tMultiple Instances are not allowed."
-        });
+
+        // Don't show dialog box if we are passing arguments to the first instance
+        const launchAppArg = process.argv.find(arg => arg.startsWith('--launch-app='));
```
Levev: I think we should probably use [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) here

src/main/main.ts
```diff
@@ -107,9 +114,62 @@ function createWindow() {
...
+// Handling pending launch - bl4ckk
+let pendingAppLaunch: string | null = null;
+let rendererReady = false;
+
+// Handler for ready renderer -bl4ckk
+ipcMain.on('winboat-ready', () => {
+    console.log('[Main] Renderer is ready');
+    rendererReady = true;
+
+    // Se c'è un lancio app in sospeso, invialo ora
```
Levev: Please keep comments English, so other people can understand as well

src/main/main.ts
```diff
@@ -107,9 +114,62 @@ function createWindow() {
...
+        mainWindow.webContents.send('launch-app-from-cli', appPath);
+    } else {
+        console.log('[Main] Renderer not ready, queuing launch');
+        pendingAppLaunch = appPath;
+    }
+}
+
+// New stuff for app creation - bl4ckk
+import * as path from 'path';
+import * as fs from 'fs';
```
Levev: Imports should be placed at the top of the file

src/main/main.ts
```diff
@@ -107,9 +114,62 @@ function createWindow() {
...
+
 app.whenReady().then(() => {
+
+    const WINBOAT_DIR = path.join(app.getPath('home'), '.winboat');
```
Levev: Constants should be placed at the top of the file as well

src/renderer/App.vue
```diff
@@ -143,8 +143,8 @@ const remote: typeof import('@electron/remote') = require('@electron/remote');
 const $router = useRouter();
 const appVer = import.meta.env.VITE_APP_VERSION;
 const isDev = import.meta.env.DEV;
-let winboat: Winboat | null;
-let wbConfig: WinboatConfig | null;
+let winboat: Winboat | null = null; // changed - bl4ckk
+let wbConfig: WinboatConfig | null = null; // changed - bl4ckk
```
Levev: If we assign `winboat` or `wbConfig` a value of null, then vue type inferance will only infer the type `null`.

```suggestion
let winboat: Winboat | null;
let wbConfig: WinboatConfig | null;
```

src/renderer/lib/winboat.ts
```diff
@@ -701,6 +701,210 @@ export class Winboat {
...
+        const guestApiUrl = `http://127.0.0.1:${GUEST_API_PORT}`;
+
+        const cleanAppName = app.Name.replace(/[,.'"]/g, "");
+        const appFile = app.Name.replace(/\s+/g, '_');
+
+        // Paths on best effort
+        const homeDir = os.homedir();
+        const binDir = path.join(homeDir, '.local', 'bin');
+        const appDir = path.join(homeDir, '.local', 'share', 'applications');
+        const iconDir = path.join(homeDir, '.local', 'share', 'icons', 'winboat');
```
Levev: Need to investigate whether this works on immutable distros

src/renderer/lib/winboat.ts
```diff
@@ -701,6 +701,210 @@ export class Winboat {
...
+        // Get credentials and port
+        const { username, password } = this.getCredentials();
+        const rdpHostPort = this.getHostPort(GUEST_RDP_PORT);
+        const freeRDPBin = await getFreeRDP();
+
+        // get stockArgs as string
+        const stockArgsStr = stockArgs.join(' ');
+
+        // Generate launcher script to dinamically read configs
+        const launcherScript = `#!/usr/bin/env bash
```
Levev: You should separate this out into it's own file

Levev (changes requested):

src/main/main.ts
Bl4ckk98: Doesnt exist anymore

src/main/main.ts
Bl4ckk98: You are totally right, also this one got deleted

src/main/main.ts
Bl4ckk98: This is also deleted stuff

src/main/main.ts
Bl4ckk98: This is also deleted stuff

src/renderer/lib/winboat.ts
Bl4ckk98: Alright... I will soon have a test on an immutable one

src/renderer/lib/winboat.ts
Bl4ckk98: May you please be more specific? I didnt get this one

src/renderer/App.vue
Bl4ckk98: Rolled back, it was to make older version working. Thank you

src/renderer/views/Apps.vue
```diff
@@ -377,6 +383,21 @@ async function removeCustomApp(app: WinApp) {
...
 
+/**
+ * Strange way to comment... anyway, this is new
+ */
+
+async function createShortcut(app: WinApp) {
+    try {
+        const result = await winboat.createDesktopShortcut(app);
+        console.log('Shortcut created:', result);
+        // Opzionale: mostra notifica di successo nell'UI
```
Bl4ckk98: Removed comment as the next one

Bl4ckk98:
I'm right know working on centralizing everything on a single script to avoid creating multiple ones (one for every shortcut).
Also creating a category for those apps.

Issue #349, opened 2025-10-15
[Feature] Front end for winget/scoop/chocolatey


Simon-Weij:
### Feature Description

Have a GUI to install apps through winget/scoop/chocolatey

### Use-Case

You won't have to install apps manually but can just search for them and install them

### Proposed Implementation

A new tab on the left called "store" or similar where you can install apps through chocolatey/scoop/winget

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [x] I would like to help out implementing this feature

upintheairsheep:
Would be a great thing, even if just winget

XSpark0:
You just need to install [UniGetUI](https://github.com/marticliment/UniGetUI) and it would show up in Apps section of WinBoat.

PR #351, opened 2025-10-15
App Store functionality

Simon-Weij:
For #349 to easily install apps from scoop/choco/winget


Simon-Weij:
From what I saw, everything runs as administrator, this makes some things really difficult. Like scoop doesn't really like this and winget isn't "installed", I'm not comfortable enough with powershell to know how to bypass this or what's causing this. Is there any reason Winboat launches most applications as Administrator?

PR #353, opened 2025-10-16
Added automatic dark mode support

adeel26in:

I have edited the file "tailwind.config.ts" to add support for Dark Mode. This enables Dark Mode not inside the Windows container but the UI/UX of the app instead. It automatically detects if you have Dark Mode and changes the colour of the app like that. This keeps it consistent across apps and also maintains user freedom. 

Kind Regards,
adeel26in


Levev:
Not sure what the point of this PR is, as WinBoat only supports Dark Mode at the moment, regardless of the DE's theming.

PR #360, opened 2025-10-17
feat: Enable Network sharing with host.docker.internal

PhantomDave:
I Was using this amazing app to do some work, but i had a legacy windows application  that needed to connect to a database, we connect to the database using an SSH Tunnel and it was not working because the container was not sharing it' s network with the Linux machine

so i added the possibility to do that,

These are really minor edits, but should be pretty self-explanatory

image

### DIFF
--- a/src/renderer/views/SetupUI.vue
+++ b/src/renderer/views/SetupUI.vue
@@ -442,24 +442,12 @@
                     <div v-if="currentStep.id === StepID.SHOULD_ADD_HOST_DOCKER_INTERNAL" class="step-block">
                         <h1 class="text-3xl font-semibold">{{ currentStep.title }}</h1>
                         <p class="text-lg text-gray-400">
-                            WinBoat allows you to share your host machine network with the Windows virtual machine, here
-                            you can choose whether to enable this feature or not.
-                        </p>
-                        <p class="text-lg text-gray-400">
-                            <b>⚠️ WARNING:</b>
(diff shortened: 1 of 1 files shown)


src/renderer/views/SetupUI.vue
PhantomDave, line 1: Seems like i had prettier enabled, let me know if you want this to be reverted

src/renderer/lib/install.ts
```diff
@@ -141,6 +141,12 @@ export class InstallManager {
         composeContent.services.windows.environment.USERNAME = this.conf.username;
         composeContent.services.windows.environment.PASSWORD = this.conf.password;
         
+        
+        if (this.conf.addHostDockerInternal) {
+            composeContent.services.windows.extra_hosts = ["host.docker.internal:host-gateway"];
```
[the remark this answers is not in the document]
PhantomDave: there aren't any extra hosts for now

src/renderer/views/SetupUI.vue
waffles-dev: Formatting with prettier is good provided it follows the .prettierrc rules - some files have likely missed being formatted.
(Our .prettierrc probably needs some improvements as well in future - but that's a separate discussion)

Levev:
WinBoat allows users to edit their compose file directly, and I think this is a great example of why that is. Adding this setting to the winboat config isn't necessary in my opinion.

I'll be closing this PR for now.

PR #367, opened 2025-10-19
Chore: Update .prettierrc and run formatter on all files

waffles-dev:
Updated .prettierrc in consultation with Tibix and Levev.
Ran on all ts, vue, html and json files - not touching xel files.

Can you please do a sanity check (builds for me, ensure builds for you)?

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -2,69 +2,69 @@ name: 🚨 Bug Report
 description: Report an issue that you've encountered
 title: "[Bug] "
 labels:
-  - bug
-  - needs-triage
+    - bug
+    - needs-triage
 body:
-  - type: markdown
-    attributes:
-      value: |
-        Thanks for taking your time reporting an issue.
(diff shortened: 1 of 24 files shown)


Levev:
What was the prettier command you ran? Running `prettier --write .` didn't work for me as it reverted a lot of the changes, as if it didn't care about the prettier rules.

waffles-dev:
> What was the prettier command you ran? Running `prettier --write .` didn't work for me as it reverted a lot of the changes, as if it didn't care about the prettier rules.

I manually ran using the prettier extension in vscode rather than working out the CLI args as it was getting late

waffles-dev:
ok, just ran with `prettier . --write` after installing via `npm i -g prettier` and it changed a few, I also had to fix an img tag in App.vue

seems the extension is slightly different to the CLI - I think we should probably rely on the CLI for ultimate truth and roll it out as part of the github workflow

waffles-dev:
added prettier to dev dependencies, after `npm i` you can run with `npx prettier . --write` and it should always run the exact same version

waffles-dev:
Just caught up with latest main, changes made in e80b9ab (just merged before) were already in this PR, so no diff to be shown

PR #369, opened 2025-10-19
Duplicate purge in tailwind.config.ts

adeel26in:

I have made a small change in the file "tailwind.config.ts" and that is removing the duplicate purge. This makes the code cleaner and doesn't affect any functionality. (This is a really small contribution but I hope you accept it :)

Kind Regards,
adeel26in


Levev:
Good catch!

PR #387, opened 2025-10-21
Fixed CPU Usage spiking when RDP Monitoring is enabled

Kuritsu243:
When the "RDP Monitoring" Option is enabled in the WinBoat GUI, the guest VM tends to suffer from high CPU usage due to the amount of times that `powershell.exe` is being instanced. I have resolved this by letting the guest server use `quser.exe`, usually returns an output such as:
```
PS C:\Users\kuritsu> quser.exe
 USERNAME              SESSIONNAME        ID  STATE   IDLE TIME  LOGON TIME
>kuritsu               rdp-tcp#0           1  Active          .  10/21/2025 12:53 PM
PS C:\Users\kuritsu>
```

I have then updated `main.go` to let the output still be returnable in json format as expected by the remainder of the program, by implementing a new struct:
```go
type StatusResponse struct {
	RdpConnection bool `json:"rdp_connection"`
}
```

and then updating the function "getRdpConnectedStatus" accordingly:
```go
func getRdpConnectedStatus(w http.ResponseWriter, r *http.Request) {
	// Check for RDP Status via quser.exe
	cmd := exec.Command("quser.exe")
	output, err := cmd.Output()
	if err != nil {
		http.Error(w, "Failed to execute script: "+err.Error(), http.StatusInternalServerError)
		return
	}
	// Check if the output contains both "active" and "rdp" todo: Check for VNC Sessions
	hasRdpSession := strings.Contains(strings.ToLower(string(output)), "active") &&
		strings.Contains(strings.ToLower(string(output)), "rdp")

	response := StatusResponse{
		RdpConnection: hasRdpSession,
	}

	// Convert the response from guest server to JSON
	// Expected output { "rdp_connected": "true" } if a session is active
	jsonResponse, err := json.Marshal(response)
	if err != nil {
		http.Error(w, "Failed to marshal JSON response: "+err.Error(), http.StatusInternalServerError)
		return
	}

	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)
	w.Write(jsonResponse)
}
```

### CPU Usage on an idle VM as it is currently:
screenshot-2025-10-21_22-42-13

### CPU Usage on an Idle VM with my method:
screenshot-2025-10-21_22-11-13

### DIFF
--- a/guest_server/main.go
+++ b/guest_server/main.go
@@ -46,7 +46,7 @@ type Metrics struct {
 }
 
 type RDPStatusResponse struct {
-	RdpConnection bool `json:"rdp_connection"`
+	RdpConnection bool `json:"rdpConnected"`
 }
 
 func getApps(w http.ResponseWriter, r *http.Request) {


guest_server/main.go
```diff
@@ -44,6 +45,10 @@ type Metrics struct {
 	} `json:"disk"`
 }
 
+type StatusResponse struct {
```
TibixDev: Rename to RDPStatusResponse

guest_server/main.go
```diff
@@ -120,17 +125,32 @@ func getMetrics(w http.ResponseWriter, r *http.Request) {
 }
 
 func getRdpConnectedStatus(w http.ResponseWriter, r *http.Request) {
-	// Grep for any active RDP sessions via qwinsta
-	// Run the PowerShell script
-	cmd := exec.Command("powershell", "-ExecutionPolicy", "Bypass", "-File", "scripts\\rdp-status.ps1")
```
TibixDev, line 125: The script itself can be deleted if it's no longer needed

guest_server/main.go
```diff
@@ -120,17 +125,32 @@ func getMetrics(w http.ResponseWriter, r *http.Request) {
...
+	// Check for RDP Status via quser.exe
+	cmd := exec.Command("quser.exe")
```
TibixDev, line 129: Does quser.exe exist in both Windows 10 and 11 by default and returns the same kind of output?

guest_server/main.go
Kuritsu243: Deleted

guest_server/main.go
Kuritsu243: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/quser

QUser is in Win10 and 11, same output. Forum threads reference it from over ten years ago so should be pretty solid on that

Kuritsu243:
Will commit the changes shortly

Kuritsu243:
Should all be ready now :)

Kuritsu243:
image

Guest server now completely mirrors the original powershell output

TibixDev:
Seems good now, thanks Kuritsu! 🎉

PR #397, opened 2025-10-23
Update Windows image to v5.12

kroese:
Should be fully backwards compatible with the current (v5.07) image, but fixes a few annoying bugs.


waffles-dev:
Thanks!

kroese:
@waffles-dev Seems like this merge was overwritten by @Levev ? Because it's back now at the old version.

waffles-dev:
Thanks, I'll get him to revert

PR #402, opened 2025-10-24
fix: Add USER_PORTS to install configuration

kroese:
When the container falls back to user-mode networking instead of kernel-mode (for example when ip_tables is not available or any other permission problem) the `USER_PORTS` variable is required to forward port 7148 to reach the Winboat Guest Server API.

### DIFF
--- a/src/renderer/data/podman.ts
+++ b/src/renderer/data/podman.ts
@@ -25,6 +25,7 @@ export const PODMAN_DEFAULT_COMPOSE: ComposeConfig = {
                 PASSWORD: "MyWindowsPassword",
                 HOME: "${HOME}",
                 LANGUAGE: "English",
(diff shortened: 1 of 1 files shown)


Levev:
Thanks for noticing this detail!

Why is it that you only included the API port in USER_PORTS? 
Should we also include the QMP port (7149) in USER_PORTS in case the container falls back to user-mode networking?

kroese:
@Levev No, because the QMP port does not need to be forwarded to the VM, but to the container itself (where QEMU is running). Port 7148 is different, because it is opened inside the VM (instead of outside like 7149).

kroese:
@Levev Can you please merge this?

Levev:
LGTM!

PR #405, opened 2025-10-24
Update tailwind.config.ts (purge to content)

adeel26in:

It seems like we are using Tailwind CSS 3.14.17 (specified in package.json) but purge seems to be depreciated and doesn't work in Tailwind anymore? I have thus updated "purge" to "content" to make it work. (Also shows depreciation warnings during building WinBoat from source)  Source: https://github.com/tailwindlabs/tailwindcss/discussions/7568

image

Got this too but is a different warning about leaking memory: 

image

2 quick questions:

1: Why are we not using Tailwind CSS 4.1, it seems to be the latest version?
2: Electron-builder has support for a .snap file. Why is that not in use?

Kind Regards,
adeel26in

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -1,6 +1,6 @@
 /** @type {import('tailwindcss').Config} */
 export default {
-    purge: ["./src/renderer/index.html", "./src/renderer/**/*.{vue,js,ts,jsx,tsx}"],
+    content: ["./src/renderer/index.html", "./src/renderer/**/*.{vue,js,ts,jsx,tsx}"],
     darkMode: false, // or 'media' or 'class'
     theme: {
         extend: {},


Levev:
We are still on 3.14.17 since Tailwind 4 introduced breaking changes and we didn't migrate yet.

TibixDev:
Hey, to answer your questions:

> 1: Why are we not using Tailwind CSS 4.1, it seems to be the latest version?

Tailwind 3 -> 4 is a breaking change and not very straightforward. We tried migrating before but a bunch of stuff broke so we decided to focus on more important things. If you're in the mood you can do the migration and let us know when it works.

> 2: Electron-builder has support for a .snap file. Why is that not in use?

We're not compatible with Flatpak / Snap for now, we need a bunch of priviliged stuff atm that you don't have access to from an isolated env. 

---

Have you verified `content` instead of `purge` works in the app as a whole?

adeel26in:
image


AppImage works (Others not tested yet as I can only really test .deb or .tar.gz now as I am on Ubuntu 24.04.03 LTS): 

A minor issue when launching AppImage that it doesn't immediately recognize the checks that it does if you have them or not. You first have to back to the MIT license screen and then it works. 

image

I just wanted to say that it says on README.md that if you build from source it would give you an AppImage and an unpacked variant but I got these under Winboat/dist/: 
image

Something related to the Snap discussion (Can't we use "classic" confinement?): 
https://snapcraft.io/docs/snap-confinement

Kind Regards,
adeel26in

TibixDev:
Hey, yeah README has to be changed since our build pipeline has new things now. As for the Tailwind changes themselves, they look good and I'm merging it, thanks!

For Snap a new issue or PR can be made where we can discuss it.

PR #406, opened 2025-10-24
Fix issue: 165: Disable Animations

### LINKED ISSUE #165: [Feature] Allow disabling animations
### Feature Description

Allow disabling all the animations in the gui

### Use-Case

When gpu accel isnt good

### Proposed Implementation

_No response_

### Notice

enXov:
## Description

This PR addresses issue #165 by adding a new configuration option to disable all animations in the GUI.

## Changes Made

- Added a new `disableAnimations` boolean configuration option in `src/renderer/lib/config.ts`
- Created a new settings toggle in the Configuration view (`src/renderer/views/Config.vue`) under the "General Settings" section
- Implemented CSS rules in `src/renderer/App.vue` that disable all animations and transitions when the setting is enabled
- Added a polling mechanism to detect config changes and dynamically apply/remove the `disable-animations` class on the document body

## Implementation Details

The feature works by:
1. Adding a global CSS class (`disable-animations`) to the body element when the setting is enabled
2. Using comprehensive CSS rules to disable all animations, transitions, and Vue transition components
3. Polling the config every 100ms to detect changes (similar to the existing `rerenderCounter` pattern in the codebase)
4. Cleaning up the interval on component unmount

## Use Case

This feature is particularly useful for users experiencing poor GPU acceleration, where animations can cause performance issues or visual glitches.

## Testing

- ✅ Verified the toggle works correctly in the Configuration view
- ✅ Confirmed all animations are disabled when the setting is enabled
- ✅ Verified animations re-enable when the setting is toggled off
- ✅ Checked that the setting persists across application restarts

Closes #165

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -201,7 +201,8 @@ onMounted(async () => {
 
     // Poll for config changes since the Proxy doesn't trigger Vue reactivity
     // This is similar to how rerenderCounter is used elsewhere in the codebase
-    let lastAnimationState = wbConfig?.config.disableAnimations;
+    // Start with undefined so that the first call will always apply the current state
+    let lastAnimationState: boolean | undefined = undefined;
     animationCheckInterval = setIntervalImmediately(() => {
         const currentState = wbConfig?.config.disableAnimations;
         if (currentState !== lastAnimationState) {


TibixDev:
Hey, good change thanks, could you use `setIntervalImmediately` and a timer of 1000ms? I think it's useless to poll every 100ms.

src/renderer/App.vue
```diff
@@ -296,4 +330,46 @@ dialog::backdrop {
...
+
+/* Disable all animations when the setting is enabled */
+body.disable-animations,
+body.disable-animations *,
+body.disable-animations *::before,
+body.disable-animations *::after {
+    animation-play-state: paused !important;
+    animation-duration: 0s !important;
+    animation-delay: 0s !important;
+    animation: none !important;
```
waffles-dev, line 338: should animation:none and transition:none be at the top of their sections, or was this ordering intentional?

src/renderer/views/Config.vue
waffles-dev, line 1: Other than CSS changes I'm happy with code quality - please run `npx prettier . --write` to format each file consistently (after making the change Tibix requested) before we eventually merge.

I did see some elements are still animating (toggle switches, the performance monitor dials) is this intentional? I assume these don't have an impact - I don't mind either way if they're left out or included.

waffles-dev (changes requested):

src/renderer/App.vue
enXov: @waffles-dev actualy your right - since `animation: none` and `transition: none` are shorthand properties that override all the individual properties, they should be at the top. i simplified the CSS by removing the redundant properties and kept only the shorthand declarations.

src/renderer/views/Config.vue
enXov: @waffles-dev I ran the `npx prettier . --write` command and saved the formatting changes.

Regarding the elements that are still moving (transition buttons, performance monitor dials), I completely agree with you, meaning I don't think they significantly impact performance. Apart from what I mentioned, I intentionally left them enabled because I wasn't sure if disabling them would break functionality or cause usability issues XD. However, if you think this would be useful, what we need to do next is very simple!

waffles-dev:
Happy with changes - thank you for making them

PR #414, opened 2025-10-25
Shutdown Container after a defined period of inactivity

Kuritsu243:
Automatically shutdown the container after a period of no RDP Sessions. User can set the length of this.

https://github.com/user-attachments/assets/2819c237-5fd6-4ed2-99cc-26d0ba22db37


Kaan88:
Having a pause option instead of a full shutdown would be a great addition as well.

samucaoo1:
We need this

PR #418, opened 2025-10-26
fix: check winboat disk space instead of its parent path

### LINKED ISSUE #404: [Bug] Check disk space of actual directory instead of parent directory

0.8.7

### Your Environment

Linux Distro: **NixOS**
Desktop Environment: **i3**
FreeRDP: not sure, can check, but irrelevant

I have `/` mounted as `tmpfs` which is only allotted 4 GB. I instead have very specific directories that are mounted to my persistent storage.

While giving Winboat the install location for a Windows VM, it appends `winboat` to the directory path that I give it. So, I chose a parent directory of `~/.local/state/` for the resulting dir to be `~/.local/state/winboat/`. But Winboat complains about disk space, which is only 3 GB remaining in `~/.local/state/` but there's >200 GB available in `~/.local/state/winboat/`.

The disk space complaint seems like a bug since why check the parent directory? I'm assuming it's because Winboat assumes the appended `winboat` dir to not exist yet so it can't really check for the amount of available disk space within it. But this breaks in my environment and would love to see a fix :)

Image

### Logs

nothing relevant

### Expected Behavior

Check the disk space of the resultant directory _after_ appending `winboat` to the install path.

### Current Behavior

Checks the parent directory's available disk space.

### Possible Solution

I'm not sure. I'm fairly certain that Winboat doesn't check the resultant directory's disk space because it assumes it to not have been created. Changing this behaviour would be a non-trivial change on expectations.

### Quality Notice

GetPsyched:
Fixes #404

I personally don't like this approach as this breaks general expectations I'd have as a user. I better like (with bias) the approach I suggested [here](https://github.com/TibixDev/winboat/issues/404#issuecomment-3448350775).

Marking this as draft until there's more discussion on the tracking issue.

### DIFF
--- a/src/renderer/views/SetupUI.vue
+++ b/src/renderer/views/SetupUI.vue
@@ -943,8 +943,22 @@ const installFolderErrors = computedAsync(async () => {
         errors.push("The selected install location is not writable");
     }
 
+    // Create /winboat
+    const winboatPath = path.join(parentPath, "winboat");
+    const isWinboatPathCreated = false;
+    try {
+        fs.mkdirSync(winboatPath, { recursive: true });
+        isWinboatPathCreated = true;
+        console.log("Created winboat directory at", winboatPath);
+    } catch (err) {
+        console.error(err);
+    }
+
     // Check if we have enough disk space
-    const diskSpace = await checkDiskSpace(parentPath);
+    const diskSpace = await checkDiskSpace(isWinboatPathCreated ? winboatPath : parentPath);
+    if (isWinboatPathCreated) {
+        await fs.rm(winboatPath, { recursive: true });
+    }
     const freeGB = Math.floor(diskSpace.free / (1024 * 1024 * 1024));
     if (freeGB < MIN_DISK_GB) {
         errors.push(


waffles-dev:
My suggestion in #404 was purely because it didn't fundamentally change the intent of that check.
I'm not wed to the suggestion - happy to change intent if Tibix or Levev approve.

src/renderer/views/SetupUI.vue
```diff
@@ -943,8 +943,22 @@ const installFolderErrors = computedAsync(async () => {
         errors.push("The selected install location is not writable");
     }
 
+    // Create /winboat
+    const winboatPath = path.join(parentPath, "winboat");
+    const isWinboatPathCreated = false;
+    try {
+        fs.mkdirSync(winboatPath, { recursive: true });
+        isWinboatPathCreated = true;
```
waffles-dev, line 951: how does this work when isWinboatPathCreated is a constant?

src/renderer/views/SetupUI.vue
```diff
@@ -943,8 +943,22 @@ const installFolderErrors = computedAsync(async () => {
...
+        console.log("Created winboat directory at", winboatPath);
+    } catch (err) {
+        console.error(err);
+    }
+
     // Check if we have enough disk space
-    const diskSpace = await checkDiskSpace(parentPath);
+    const diskSpace = await checkDiskSpace(isWinboatPathCreated ? winboatPath : parentPath);
+    if (isWinboatPathCreated) {
+        await fs.rm(winboatPath, { recursive: true });
```
waffles-dev, line 960: Error: Object literal may only specify known properties, and 'recursive' does not exist in type 'NoParamCallback`

GetPsyched: Interesting; where are you seeing this error? I had a lot of trouble seeing _any_ errors. I had to rely solely on if things worked or not.

src/renderer/views/SetupUI.vue
GetPsyched: Whoops. That bool was a last-minute thing so it went untested

PR #327, opened 2025-10-13
Add support for different containerization runtimes

Levev:
This PR aims to add support for different containerization runtimes, while also implementing a podman-based solution as well.
This is very much in WIP, at time time this PR is created only the setup part is migrated, and podman is outright not implemented except for checking the prerequisites.

PR #424, opened 2025-10-27
Update port bindings to localhost in install.ts

bloopybae:
This patch limits the WinBoat Docker ports to 127.0.0.1 instead of 0.0.0.0,
preventing unintentional remote access to RDP, QMP, and web interfaces.

### DIFF
--- a/src/renderer/utils/port.ts
+++ b/src/renderer/utils/port.ts
@@ -22,18 +22,34 @@ export class ComposePortEntry extends String {
     guestPort: number;
     protocol: PortEntryProtocol = undefined;
 
-    constructor(entry: string) {
-        super(entry);
+constructor(entry: string) {
+    super(entry);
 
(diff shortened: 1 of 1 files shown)


bloopybae:
PortManager.parseCompose() didn’t handle three-part port strings (e.g. 127.0.0.1:8006:8006). Added a fix in ComposePortEntry to support both 2- and 3-part formats.
Tested successfully: installer now completes past “Creating Compose File...”

TheMichix (approved):
important security issue!!!

Levev:
Thanks for your time with this fix, we really appreciate it!
Sadly this issue already got fixed in #327.

To avoid such cases in the future, please go through our [Contribution Guidelines](github.com/TibixDev/winboat/blob/main/CONTRIBUTING.md).
We'd rather help contributors out on what they could help with, than to have a PR not make it :(

Issue #235, opened 2025-10-07
[Bug] passwords are stored in plain text in winboat host


aquast:
### What version of WinBoat are you running?

Beta v0.8.5 Prod

### Your Environment

OpenSuse Tumbleweed
KDE with wayland
FreeRDP from Distro

### Steps to Reproduce / Context

Reading winboat.log at /home/user/.winboat/winboat.log with 

`tail -f /home/user/.winboat/winboat.log`

### Logs

``` 
         "Name": "🖥 Browser Display",
        "Source": "internal",
        "Path": "http://127.0.0.0:8006",
        "Usage": 0
    }
]
2025-10-06 16:09:31 | INFO | Launching app: ⚙ Windows Desktop at path WINDOWS_DESKTOP
2025-10-06 16:09:31 | INFO | Using FreeRDP Command: 'xfreerdp'
2025-10-06 16:09:31 | INFO | Launch command:
xfreerdp /u:"user" /p:"myPlainPassword" /v:127.0.0.1 /port:3389 /cert:ignore +clipboard +f /sound:sys:pulse /microphone:sys:pulse /floatbar /scale:100 /compression &
2025-10-07 19:56:30 | INFO | Final USB database file path: /usr/share/hwdata/usb.ids
2025-10-07 19:58:24 | INFO | Starting WinBoat container...
2025-10-07 19:58:25 | INFO | Container response: WinBoat
```

### Expected Behavior

Passwords should be stored encrypted and never displayed in log files.

### Current Behavior

I learned from winboat.log winboat is calling FreeRDP with password in plain text. Additionally it displays the password as plain text in the log.

In conclusion this means the password is stored in plain text too. 

### Possible Solution

Store passwords encrypted and call FreeRDP with an encrypted password

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

sbuller:
This should really be treated as a service account, like for a database. Asking the user for a password in the first place seems like a mistake. The RDP password should be an implementation detail. Encryption isn't necessary then, just suitable permissions.

ourichermath:
I'd also like to mention that container.log also contains the plain text password.

TibixDev:
Not addressable for the compose file, the password needs to be present there. I think by far the biggest issue people had is that it was display in the logs for FreeRDP app launches, and they accidentally pasted it, I decided to change that. I think if someone wants to find your password they easily can anyway if they have local priviliges.

RDP log fixes should be in 13ec434

[...]

PR #383, opened 2025-10-21
Remove plain password from logs

gabrielecabrini:
This closes https://github.com/TibixDev/winboat/issues/235

PR #426, opened 2025-10-27
Default to randomised password and warn user about password being stored in plaintext

auri-f5bde6:
Partially solve #235

### DIFF
--- a/src/renderer/lib/password.ts
+++ b/src/renderer/lib/password.ts
@@ -1,4 +1,4 @@
-import wordsListPath from "word-list";
+const wordList: typeof import("word-list") = require("word-list");
 
 const fs: typeof import("fs") = require("node:fs");
 const crypto: typeof import("crypto") = require("node:crypto");
@@ -21,6 +21,6 @@ function getRandomWordFrom(words: string[]) {
 }
 
 export function readWordList(): string[] {
-    const wordArray = fs.readFileSync(wordsListPath, "utf8").split("\n");
+    const wordArray = fs.readFileSync(wordList.default, "utf8").split("\n");
     return wordArray.filter(word => word.length >= 5);
 }


auri-f5bde6:
> I don't think requiring "word-list" package is necessary. You could simply generate random characters by randomize number between 97-122 then convert that ascii value to a char `String.fromCharCode(ascii_code)`

I was told by @TibixDev to use the npm package, but I do somewhat agree (that's why I had the eff text list originally).
As for the second point, I opted for word/memorisable password, just in the odd case where a user might want to log into their VM though ways other than winboat. In that case, memorisable password would be a lot easier to type and remeber.

DamienStaebler:
> > I don't think requiring "word-list" package is necessary. You could simply generate random characters by randomize number between 97-122 then convert that ascii value to a char `String.fromCharCode(ascii_code)`
> 
> I was told by @TibixDev to use the npm package, but I do somewhat agree (that's why I had the eff text list originally). As for the second point, I opted for word/memorisable password, just in the odd case where a user might want to log into their VM though ways other than winboat. In that case, memorisable password would be a lot easier to type and remeber.

How about instead of word, generate random 6 digits?

auri-f5bde6:
Or we can just default the password to `winboat`, which achieves the same goal. _But_, there's also a possibility the user may wanna remotely connect to the VM, in which case, having a reasonably safe password would be a lot better as a default.

DamienStaebler:
True,  in that cause then user should choose their own password. Have a default password "winboat", give user warning and understand the risk. And give them the option to change default password to something else.

auri-f5bde6:
p.s. just realised the program doesn't even run with the npm packages, not sure how I missed that

auri-f5bde6:
> True, in that cause then user should choose their own password. Have a default password "winboat", give user warning and understand the risk. And give them the option to change default password to something else.

In my opinion, memorable password is a good middle ground, but, perhaps. I just think the default should be reasonably safe, just in case if the user didn't read properly.

auri-f5bde6:
But yeah! Thanks for the feedback, I'll have a think about it!

gabrielecabrini:
If we merge https://github.com/TibixDev/winboat/pull/383 there's no need to warn user about plaintext password

auri-f5bde6:
> If we merge https://github.com/TibixDev/winboat/pull/383 there's no need to warn user about plaintext password

Hmm, but in my opinion, it doesn't change the fact that it's stored in plaintext in `compose.yml` regardless, so there should be a warning still.

auri-f5bde6:
[Thanks](https://github.com/TibixDev/winboat/pull/436#issuecomment-3985605804)

PR #436, opened 2025-10-28
Add a new section in setting for logs (fix #375)

### LINKED ISSUE #375: [Feature] Read winboat.log file and docker logs from the app
### Feature Description

Hi,

this is a minor issue but it would be really nice to have access to winboat.log from the app, especially for non tech people. Having a "open log in file explorer" would be nice too. (one or the other or both)

I intend to install winboat for some friends that are completely obnoxious to terminal, dot files, log files ... and it would be much easier for me to ask them "click there, copy the text and send me". 

by the way : thank you for the work done here it's amazing.

### Use-Case

Installed winboat on a friend's laptop, everything was fine.
The day after, things would not start.
Reason : not enough RAM free at the moment.
This error is only available in docker logs, so it requires at least knowledge to "docker ps -a" and "docker logs".

### Proposed Implementation

From Home where containers are displayed.
Add a "winboat logs" button that opens file explorer to log file location.
Add "container logs" button that shows the latest container's logs. If no container is running, show latest one (if available)

### Notice

auri-f5bde6:


src/renderer/views/Config.vue
```diff
@@ -537,6 +537,65 @@
                         ></x-switch>
                     </div>
                 </x-card>
+
+                <!-- Logs -->
+                <x-card
```
waffles-dev: Can you please add closing of the log dialog via escape key press and clicking outside of the dialog?
Maybe a small/standard close button top right instead of across bottom would work better? Colour-wise it all blended in and I couldn't figure out how to close at first (I'm not very observant).

We don't really have many popups in winboat that I can think of off the top of my head, but maybe make the layout the same as the popup you get when the guest server is out of date?

src/renderer/views/Config.vue
```diff
@@ -537,6 +537,65 @@
...
+                            <Icon class="inline-flex text-violet-400 size-8" icon="lucide:file-code"></Icon>
+                            <h1 class="my-0 text-lg font-semibold">Logs</h1>
+                        </div>
+                        <p class="text-neutral-400 text-[0.9rem] !pt-0 !mt-0">Select the log you want to open</p>
+                    </div>
+                    <div class="flex flex-col gap-2 justify-center items-center" style="font-size: 0.8125rem">
+                        <x-button
+                            @click="
+                                logTitle = 'Winboat log (winboat.log)';
+                                logContent = fs.readFileSync(path.join(WINBOAT_DIR, 'winboat.log'));
```
waffles-dev: `Type 'NonSharedBuffer' is not assignable to type 'string'`

src/renderer/views/Config.vue
auri-f5bde6: Ok!

src/renderer/views/Config.vue
auri-f5bde6: Strange, I never had that warning. I'll have a look!

src/renderer/views/Config.vue
auri-f5bde6: Sorry, I mistakenly marked this as resolved after implementing the esc to close. But, I suppose I can use `<x-titlebar>` for the close button, but I feel like that may be a bit confusing, as the dialog is not actually a new window, and cannot be moved. Thanks!

auri-f5bde6: `<x-titlebar>` actually have no way of hiding to maximise and minimize button, annoyingly.  (at least afaik)

auri-f5bde6:
Feel free to reopen this if there's any interest to do so, but I need some closure for the time being. Thanks

PR #439, opened 2025-10-28
Preparation for fetching ports from containerization runtime

Levev:
This PR introduces changes in preparation for a better port solution.
PortManager manually edited the compose itself, while not even complying with the spec.
We should instead specify a host port range in the compose, then fetch the actual host ports from the containerization runtime using the following:
```terminal
$ docker port WinBoat_Docker
$ podman port WinBoat_Podman
```


TibixDev:
LGTM!

PR #447, opened 2025-10-30
flake: Replace freerdp3 with freerdp in nativeBuildInputs

bbigras:
freerdp3 is deprecated in recent nixpkgs:

```
error: 'freerdp3' has been renamed to/replaced by 'freerdp'
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,7 +22,7 @@
         
         nativeBuildInputs = [
           pkgs.makeWrapper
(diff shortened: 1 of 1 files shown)


waffles-dev:
We have another PR up, #434, which intends to remove this file.
Is it still necessary to keep and maintain the flake vs nixpkgs?

waffles-dev:
nix flake has been removed in #434

PR #456, opened 2025-11-04
Add tls rdp option to skip Kerberos

Bl4ckk98:
By adding this:
- the process of opening apps is actually faster
- audio, clipboard and fullscreen are still supported
- kerberos is being skipped, avoiding stuff like -- [26769:00006893] [ERROR][com.winpr.sspi.Kerberos] - [kerberos_AcquireCredentialsHandleA]: krb5glue_get_init_creds (Client's credentials have been revoked [-1765328366]) -- which currently delays the process of opening apps cause of rdp falling back to ntlm only after kerberos. It doesnt in fact require kerberos or a domain

### DIFF
--- a/src/renderer/lib/winboat.ts
+++ b/src/renderer/lib/winboat.ts
@@ -76,6 +76,7 @@ const stockArgs = [
     "/microphone:sys:pulse",
     "/floatbar",
     "/compression",
+    "/sec:tls",
 ];
 
 /**


Bl4ckk98:
As a reference for [why I made this](https://discord.com/channels/1408451891523752017/1408455996854239365/1435268768539344937)

Levev:
LGTM!

PR #465, opened 2025-11-09
Favorite app

DamienStaebler:
This PR aims to provide a way to pin/favorite app to the top, making it easier to find without having to scroll through big app list.

image

Bit extra, added a way to copy running command to clipboard.
image


DamienStaebler:
I forgot to pull from upstream

PR #506, opened 2025-11-22
fix: Add default homebrew bin directory to process.env.PATH

Levev:
A big chunk of our userbase uses WinBoat under an immutable distro through GearLever.
In case it's the flatpak version of GearLever, PATH, and some other environment variable will get stripped by default.
We include the default homebrew bin directiory for exactly this reason.
It's not WinBoat's responsibility if the PATH envvar is incomplete, but in this case it affects a lot of users.


TibixDev:
LGTM! 🙌

PR #508, opened 2025-11-22
feat: Add version data to config

Levev:
This way we could not only gain useful information for debugging purposes, but we could also implement better migrations that rely on knowing what version the user is updating from.


Levev:
Though we'll have to separate wbConfig into wbData and wbConfig later down the line so as to respect XDG spec.

TibixDev:
LGTM! 🙌

PR #512, opened 2025-11-23
feat: Better auto migrations, Add migration step for pre-0.9.0

Levev:
With this we are in the finish line for getting 0.9.0 out (hopefully)


TibixDev:
LGTM! 🙌

PR #513, opened 2025-11-23
chore: Add loading animation in place of novnc while the container is starting

Levev:
This is so users don't accidentally open the wrong link


TibixDev:
LGTM! 🙌

PR #525, opened 2025-11-26
feat: auto install requirements

AdianKozlica:
Here is a video showcasing the process:

[autosetup3.webm](https://github.com/user-attachments/assets/b6f6e7f7-b945-435a-9217-147058ceb348)

### DIFF
--- a/autosetup/autosetup_apparmor.sh
+++ b/autosetup/autosetup_apparmor.sh
@@ -7,5 +7,6 @@ if [ -f /etc/apparmor.d/local/usr.sbin.smbd ]; then
 /shared/ lrwk,
 /shared/** lrwk,
 EOF
+"$ESCALATION_TOOL" apparmor_parser -r /etc/apparmor.d/usr.sbin.smbd
 printf "%b\n" "${GREEN}SMBD Apparmor profile modified!${RC}"
 fi


TibixDev:
Very cool idea, it'd be helpful for users if the setup was automatic, but there's simply just tons of distros.
(I encourage you to include the video here that you left on Discord showcasing the process)

I think this needs tons of visual polish, here's a non-exhaustive list:
- Needs animations
- It shouldn't be called "autosetup" since it can be confused with the WinBoat setup itself
- Button placement should be different
- The option for autosetup should only appear with a compatible distro
- Terminal component needs backdrop
- The autosetup page needs some description on what it does and how (We'll attempt to install the dependencies for you, you can follow the process here - here's roughly what will happen)
- The sudo prompt shouldn't happen before you enter the autosetup, since the user might be confused
- There has to be a failure mode where the script didn't succeed
- In case of Docker there's at least 1 step with the group add where the user needs to relog, needs to be indicated

The feature also needs extensive testing on supported distros.

ShaunRoselt:
Very cool!

Please can you test and make sure it works on [SteamOS](https://store.steampowered.com/steamos) also 🙏🏻

It's so much effort to install those dependencies on SteamOS. It would be nice if this can be automated!

Issue #558, opened 2025-12-08
[Feature] RDP Performance Tweaks


nukhes:
### Feature Description

RDP can be configured to offer better performance and even achieve a frame rate close to 60 FPS.

I tested various of registry keys and selected the one that truly improves the experience based in the following sources:
https://www.reddit.com/r/killerinstinct/comments/4fcdhy/an_excellent_guide_to_optimizing_your_windows_10/
https://learn.microsoft.com/en-us/troubleshoot/windows-server/remote/frame-rate-limited-to-30-fps

### Use-Case

The performance improvement, at least on my machine running Arch Linux, was noticeable, reaching an 60FPS-like experience.

### Proposed Implementation

You only need to add a reg file with the changes and import it in the install.bat file on the guest machine.

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [x] I would like to help out implementing this feature

PR #559, opened 2025-12-08
feat: Add RDP performance tweaks registry settings

nukhes:
issue: #558

### DIFF
--- a/guest_server/RDPTweaks.reg
+++ b/guest_server/RDPTweaks.reg
@@ -1,0 +1,31 @@
+Windows Registry Editor Version 5.00
+
+    ; Additional RDP performance tweaks
+    [HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Terminal Server\WinStations]
+    "DWMFRAMEINTERVAL"=dword:0000000f
+
+    [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Dwm]
+    "OverlayMinFPS"=dword:00000000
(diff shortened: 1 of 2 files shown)


B4T3S:
Tested this config on my machine (Thinkpad E14 with Fedora 42) and it made a *worlds* difference!

TibixDev:
So for me this tweak makes it much snappier and smoother, but @Levev for example had very weird lags and artifacts after it, so it still remains to be tested by others, feedback welcome.

BPplays:
`DWMFRAMEINTERVAL` seems to be the minimum frametime for RDP, i tested with it set to 8 and using 
`Get-WmiObject win32_videocontroller | fl CurrentRefreshRate,MaxRefreshRate,MinRefreshRate` seems to support that assumption: it reported the max framerate for the RDP device as 125.  

so `DWMFRAMEINTERVAL` would ideally be set based on the users display not just always for 60fps.
that would require setting the registry every time the user boots the container, and might lag by 1 boot, so it might be better to just set it high by default.
i'm not sure if freeRDP does a good job auto negotiating based on the clients framerate so im not sure if it would always try to run at max

Quil180:
Just tested it on nixos on my laptop after fiddling with a shell.nix for an hour. RDP runs much smoother and is very noticable. Hopefully it can be merged soon into main!

evsl:
Same experience - also NixOS user. Much, **much** smoother performance!

Stensel8:
Any updates on this?

PR #576, opened 2025-12-16
feat: Add custom folder mounts configuration

### LINKED ISSUE #44: Extra bind mounts in docker-compose
Hi and thank you for some really cool software!

I was wondering if today we can mount more paths in the compose file and they can be accessed in the VM in any way?
I was trying on use winboat to run some modding tools for games (not run the games, only modifying them) and would have need for something like this:
```
    volumes:
      - "data:/storage"
      - "${HOME}:/shared"
      - "./oem:/oem"
      - "/mnt/games:/games"
```
As windows did not seem to like softlinks, which I tried to setup inside my homefolder in linux, and hardlinks does not work with folders.
Any help appreciated.

### LINKED ISSUE #343: [Feature] Add ability to mount another folder from host
### Feature Description

Instead of just mounting the `home/` directory, it would be good to be able to mount another directory or another mount from the host machine.

### Use-Case

I have a shared drive in my desktop that is shared between Windows and Linux and contains data that is needed for an app that I use that is Windows only. Currently, I don't have a good way to map this shared data to somewhere on my home directory so that it's accessible to the app that runs in Winboat.

### Proposed Implementation

_No response_

### Notice

W1talka:
## Summary

  Add the ability for users to mount custom folders from the Linux host
  filesystem into the Windows VM. Mounted folders appear in Windows at
  `\\host.lan\Data\<name>`.

  **Features:**
  - Add/remove custom folder mounts via the Configuration panel
  - Browse for folders using native file picker
  - Enable/disable individual mounts without removing them
  - Path validation (checks directory exists and is accessible)
  - Share name validation (alphanumeric, underscores, hyphens only)

  ## Changes

  - `CustomVolumeMounts.vue` - New component for managing mounts
  - `volumes.ts` - Validation helpers and compose config utilities
  - `types.ts` - New `CustomVolumeMount` type
  - `Config.vue` - Integration into settings page
  - `config.ts` / `winboat.ts` - Backend integration

  ## Test Plan

  - [x] Add a new mount via folder browser
  - [x] Add a mount by typing path manually
  - [x] Verify validation errors appear for invalid paths
  - [x] Toggle mount on/off
  - [x] Remove a mount
  - [x] Restart WinBoat and verify mounts persist
  - [x] Access mounted folder from Windows at `\\host.lan\Data\<name>`

  Closes #343
  Closes #44

### DIFF
--- a/src/renderer/lib/volumes.ts
+++ b/src/renderer/lib/volumes.ts
@@ -4,44 +4,32 @@ const path: typeof import("path") = require("node:path");
 import type { CustomVolumeMount, ComposeConfig } from "../../types";
 
 /**
- * Validates a host path exists and is accessible
+ * Validates a host path exists and is accessible.
+ * Throws an Error if the path is invalid.
  */
-export function validateHostPath(hostPath: string): { valid: boolean; error?: string } {
+export function validateHostPath(hostPath: string): void {
+    if (!hostPath) throw new Error("Path is required");
+    if (!path.isAbsolute(hostPath)) throw new Error(`Path must be absolute: '${hostPath}'`);
+    if (!fs.existsSync(hostPath)) throw new Error(`Path does not exist: '${hostPath}'`);
+
+    const stats = fs.statSync(hostPath);
+    if (!stats.isDirectory()) throw new Error(`Path is not a directory: '${hostPath}'`);
+
     try {
-        if (!hostPath) {
-            return { valid: false, error: "Path is required" };
-        }
-        if (!path.isAbsolute(hostPath)) {
-            return { valid: false, error: "Path must be absolute" };
-        }
-        if (!fs.existsSync(hostPath)) {
-            return { valid: false, error: "Path does not exist" };
-        }
-        const stats = fs.statSync(hostPath);
-        if (!stats.isDirectory()) {
-            return { valid: false, error: "Path is not a directory" };
-        }
         fs.accessSync(hostPath, fs.constants.R_OK);
-        return { valid: true };
     } catch {
-        return { valid: false, error: "Path is not accessible" };
+        throw new Error(`Path is not accessible: '${hostPath}'`);
     }
 }
 
 /**
- * Validates a share name (folder name visible in Windows)
+ * Validates a share name (folder name visible in Windows).
+ * Throws an Error if the name is invalid.
  */
-export function validateShareName(name: string): { valid: boolean; error?: string } {
-    if (!name) {
-        return { valid: false, error: "Name is required" };
-    }
-    if (!/^[a-zA-Z0-9_-]+$/.test(name)) {
-        return { valid: false, error: "Only letters, numbers, underscores, hyphens" };
-    }
-    if (name.length > 32) {
-        return { valid: false, error: "Max 32 characters" };
-    }
-    return { valid: true };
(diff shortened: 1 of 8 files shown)


src/renderer/lib/volumes.ts
```diff
@@ -0,0 +1,126 @@
+const fs: typeof import("fs") = require("node:fs");
+const path: typeof import("path") = require("node:path");
+
+import type { CustomVolumeMount, ComposeConfig } from "../../types";
+
+/**
+ * Validates a host path exists and is accessible
+ */
+export function validateHostPath(hostPath: string): { valid: boolean; error?: string } {
```
Levev: This is a very interesting way to handle errors/perform validation, and it differs quite a bit from how the rest of the codebase goes about it. 

ComposePortEntry parses short form port mappings, you should check out how it handles this.

src/renderer/lib/winboat.ts
```diff
@@ -442,6 +445,32 @@ export class Winboat {
...
+        if (commands.length === 0) {
+            logger.info("No custom volume mounts to symlink");
+            return;
+        }
+
+        logger.info(`Creating symlinks for ${commands.length} custom volume mount(s)`);
+
+        for (const cmd of commands) {
+            try {
+                await execAsync(`docker exec WinBoat ${cmd}`);
```
Levev: This won't work with podman, you should use `this.containerMgr.executableAlias` instead.

Levev (changes requested):

src/renderer/lib/volumes.ts
W1talka: Switched to throw-based validation matching ComposePortEntry's pattern. Error messages now include contextual values (e.g. Path must be absolute: 'foo/bar'). Tested all validation paths manually.

src/renderer/lib/winboat.ts
W1talka: Replaced with this.containerMgr!.executableAlias. Verified on a fresh Podman install — symlinks created successfully and mount accessible in Windows at \\host.lan\Data\<name>.

soultaco83:
Seems if you have "Shared Home Folder" and "Custom Folder Mounts" enabled. Only the home folder will appear. If I disabled the Shared Home Folder the custom folder mounts and can be located.

W1talka:
Tested manually on both Docker and Podman.

  Docker:
  - custom mount only: visible under \\host.lan\Data2\<name>
  - custom mount + home share: home stays under Data, custom mount stays
  under Data2
  - disable custom mount: custom mount disappears
  - disable home share: nothing remains
  - re-enable custom mount: custom mount returns under Data2
  - multiple custom mounts: all appear under Data2
  - persistence: after full app restart, mounts still appear correctly

  Podman:
  - home share + custom mount: both visible at the same time
  - custom only: Data empty, Data2 shows the custom mount
  - home only: Data shows home share, no Data2 custom mount
  - nothing shared: neither Data nor Data2 shows shared content
  - custom re-enable: Data2 custom mount returns correctly
  - home share + 3 custom mounts: all 3 appear under Data2 and content is
  accessible

TopchetoEU:
sorry to bother, is there any progress on this? having this feature would be very convenient...

Issue #308, opened 2025-10-11
[Bug] When running on native wayland, UI laggs a lot


d1BG:
### What version of WinBoat are you running?

installed on 0.8.5 (cachy repos), updated to 0.8.7 (aur)

### Your Environment

Linux Distro used: Arch Linux
Desktop Environment used: Niri 25.08 (01be0e6)
FreeRDP type and version used: 3.17.2

Im running on an RTX 3060 12GB GPU (which might be the cause).
The lag happens when using env var `ELECTRON_OZONE_PLATFORM_HINT="wayland"`

The VM and apps themself are okay, its just the Winboat UI that lags.

### Steps to Reproduce / Context

1. Set `ELECTRON_OZONE_PLATFORM_HINT="wayland"` (in my case its set globally)
2. Launch WinBoat
3. UI is really laggy

### Logs

[install.log](https://github.com/user-attachments/files/22866132/install.log)
[winboat.log](https://github.com/user-attachments/files/22866131/winboat.log) 

I dont think either of them can help a lot tho..

### Expected Behavior

The UI should run smooth.

### Current Behavior

The UI laggs like crazy, sometimes unusable.

### Possible Solution

A temporary 'fix' for me is to run with `ELECTRON_OZONE_PLATFORM_HINT="x11"` - not ideal since X11 is insecure. In addition Niri doesnt have an x11 server of its own, so something like `xwayland-satellite` is needed.

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

d1BG:
Additional Stuff that i should have posted, Heres what it shows in terminal when i launch from there:
```
[68535:1011/221217.872961:ERROR:gbm_pixmap_wayland.cc(82)] Cannot create bo with format= RGBA_8888 and usage=SCANOUT
[68535:1011/221217.873009:ERROR:gbm_pixmap_wayland.cc(82)] Cannot create bo with format= RGBA_8888 and usage=GPU_READ
[68535:1011/221217.873036:ERROR:shared_image_factory.cc(898)] CreateSharedImage: could not create backing.
[68535:1011/221217.873064:ERROR:gbm_pixmap_wayland.cc(82)] Cannot create bo with format= RGBA_8888 and usage=SCANOUT
[68535:1011/221217.873081:ERROR:gbm_pixmap_wayland.cc(82)] Cannot create bo with format= RGBA_8888 and usage=GPU_READ
[...]

PR #581, opened 2025-12-18
chore: bump electron to v39

rupansh:
fixes https://github.com/TibixDev/winboat/issues/308

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
                 "autoprefixer": "^10.4.20",
                 "chalk": "^4.1.2",
                 "chokidar": "^3.6.0",
-                "electron": "^35.1.5",
+                "electron": "^39.2.7",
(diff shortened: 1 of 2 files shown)


TibixDev:
Looks good, however have you checked if this doesn't cause any problems on slightly older distros?

rupansh:
> Looks good, however have you checked if this doesn't cause any problems on slightly older distros?

I am currently running an up-to-date arch linux machine.
I checked https://www.electronjs.org/docs/latest/breaking-changes and there doesn't seem to be anything big

I can try running on an older distro of your preference however

TibixDev:
Hey, sorry for the delay, but yeah you're right, and Electron 39 has been out for a while so it shouldn't cause any major issues if it didn't in other apps to date.

LGTM! 🙌

PR #598, opened 2025-12-23
Removed Skype.

mishastarrrr-ctrl:
Skype has reached its end of life.


TibixDev:
Hey, this feature isn't really implemented nor is the list finished, just left there as a concept, but if we do get around it we'll remove Skype.

PR #618, opened 2026-01-02
Copying a file leaves the original perms intact

orkwitzel:
Changed the copyFile function in the guest_server, so now it wouldn't let the user surpass the original app perms and mode by copying it. Leaves the original perms intact.


TibixDev:
This is a good change, better than what we had, it's good to keep permissions in-tact.

LGTM! 🙌

PR #620, opened 2026-01-02
Add usb port passthrough (hostbus+hostport)

iitzrohan:
This PR adds USB Port Passthrough as an alternative to the existing USB Device Passthrough. Instead of tracking devices by vendorId:productId, it tracks them by bus + port combination. This utilizes most of the already working code used in USB device passthrough.

The existing USB passthrough works great for most cases, but fails when a device's ID changes mid-session. A common example is firmware updates - when you update a controller's firmware, it often re-enumerates with a different VID:PID (e.g., entering DFU/bootloader mode), causing it to disconnect from the VM.

I tested my Cyclone 2 firmware update using this passthrough in Xbox wired mode. It changes IDs during the process, so device passthrough was not an option.

This uses hostbus + hostport combination to passthrough devices. When a device connects to a registered port, it automatically gets passed through to the VM regardless of its VID:PID.

Note:

Port passthrough doesn't work when you attach a USB 3.x device after the port was registered with a USB 2.0 device (or vice-versa). This is because USB 3.0 ports have separate controllers for USB 2.0 and USB 3.0 protocols, resulting in different bus numbers for the same physical port.


TYeniyayla:
Can you enable bidirectional operation? That is, can you transfer a USB drive from a virtual machine to the host machine? I mentioned this in the previous issue: [#issue-3835503330](https://github.com/TibixDev/winboat/issues/647#issue-3835503330)

PR #649, opened 2026-01-22
add sdl-freerdp3 as a valid RDP client choice

### LINKED ISSUE #646: [Feature] winboat use wayland freerdp when available
### Feature Description

Most desktop environments have migrated to wayland, and X is being removed from many distributions this year.  winboat looks for "xfreerdp3" but I can only use wlfreerdp3 on my ubuntu 26.04.

The pre-requisite check when I started winboat would fail, it would never find freerdp.
I looked at the source code, and saw that it was looking for *xfreerdp3*, *xfreerdp*, or the flatpack version.
so I did a symbolic link:
```shell

% cd /usr/bin
% ln -s wlfreerdp3 xfreerdp3

```
And then the pre-requisite detection worked, and I was able to successfully install windows and run it.

Looking at the source code, it looks like one should just add another line, say here:
```
const freeRDPInstallations = [
    new FreeRDPInstallation("xfreerdp3"),
    new FreeRDPInstallation("xfreerdp"),
    new FreeRDPInstallation("flatpak", ["run", "--command=xfreerdp", "com.freerdp.FreeRDP"]),
];

```

should be changed to add the wayland implementations as well.

### Use-Case

Any linux system with a wayland desktop should be using wlfreerdp3 instead of X.  It avoids having to set up x under wayland.  Using x clients will get progressively more andiquated complicated over time, as wayland is now a lot more popular.

### Proposed Implementation

The stanza above should become something like:

```
const freeRDPInstallations = [
    new FreeRDPInstallation("xfreerdp3"),
    new FreeRDPInstallation("xfreerdp"),
    new FreeRDPInstallation("flatpak", ["run", "--command=xfreerdp", "com.freerdp.FreeRDP"]),
    new FreeRDPInstallation("wlfreerdp3"),
    new FreeRDPInstallation("flatpak", ["run", "--command=wlfreerdp", "com.freerdp.FreeRDP"]),
];

```

It would still prefer x one if present, but at least detect the wayland one if the x one isn't there.

### Notice

petersilva:
close #646 

as per discussion in that issue: 

* I don't think we should check for sdl-freerdp3 on older systems:  It's unlikely to be better than X on them.
* I don't think we need to worry about xfreerdp3 on newer systems, it will likely not be present.  If it is present, it is highly likely that it should be removed, so even if the error shows up, it's a real error that the user should fix (by removing it.) so it's good information, and easily addressed.
* I don't think checking for presence of wayland is enough to make the decision to use it, and I don't think re-implementing logic already present in SDL itself is easy or helpful.

but those are all opinions... everybody can have different ideas... I made a PR in case someone thinks it's useful, or not... 

(copied on request from @Hi-Angel )

### DIFF
--- a/src/renderer/utils/getFreeRDP.ts
+++ b/src/renderer/utils/getFreeRDP.ts
@@ -25,6 +25,7 @@ const freeRDPInstallations = [
     new FreeRDPInstallation("xfreerdp3"),
     new FreeRDPInstallation("xfreerdp"),
     new FreeRDPInstallation("flatpak", ["run", "--command=xfreerdp", "com.freerdp.FreeRDP"]),
+    new FreeRDPInstallation("sdl-freerdp3"),
 ];
 
 /**
--- a/src/renderer/utils/getFreeRDP.ts
+++ b/src/renderer/utils/getFreeRDP.ts
@@ -26,6 +26,9 @@ const freeRDPInstallations = [
     new FreeRDPInstallation("xfreerdp"),
     new FreeRDPInstallation("flatpak", ["run", "--command=xfreerdp", "com.freerdp.FreeRDP"]),
     new FreeRDPInstallation("sdl-freerdp3"),
+    new FreeRDPInstallation("sdl-freerdp"),
+    new FreeRDPInstallation("sdl3-freerdp3"),
+    new FreeRDPInstallation("sdl3-freerdp"),
 ];
 
 /**


src/renderer/utils/getFreeRDP.ts
```diff
@@ -25,6 +25,7 @@ const freeRDPInstallations = [
     new FreeRDPInstallation("xfreerdp3"),
     new FreeRDPInstallation("xfreerdp"),
     new FreeRDPInstallation("flatpak", ["run", "--command=xfreerdp", "com.freerdp.FreeRDP"]),
+    new FreeRDPInstallation("sdl-freerdp3"),
```
Hi-Angel, line 28: But why not detect `sdl-freerdp3` on Wayland first, as was discussed? On Wayland systems it presumably should have better performance.

petersilva: If you´re on Wayland, the X ones should not work... barring having done some compatibility voodoo.  Binaries built for X won't run unless some sort of adaptation software is present.  Since the x-based binaries won't work, there is no reason for them to be present.  In fact, there is a good reason for them to be removed.

If you want to make the choice of which binary to check based on the display server, then you are back to re-implementing detection, and making a choice of which one to use, because the availability of Wayland (or X) doesn´t mean that´s the display server you want to use.  You then need new ways to override one way or the other if both are present.  Both being present is a very common case on systems for the last few years.

Based on that logic alone, the *FreeRDPInstallation* checks could be placed in any order.  But since we know that sdl-freerdp3 performs badly on some older installations,  looking for sdl- version afterwards should minimize the number of issues on older systems.

That was my reasoning... it might not be right though.

Hi-Angel: > If you´re on Wayland, the X ones should not work...

Well, you detect those here first… Anyway, I don't think it's a problem, because I doubt people would start removing XWayland from their Wayland systems for another decade at least *(due to older games/apps that just won't get ported)*.

> If you want to make the choice of which binary to check based on the display server, then you are back to re-implementing detection, and making a choice of which one to use, because the availability of Wayland (or X) doesn´t mean that´s the display server you want to use. You then need new ways to override one way or the other if both are present. Both being present is a very common case on systems for the last few years.

I don't see anything complicated, you just check if `$WAYLAND_DISPLAY` is set. If it is, that means we're running Wayland, so try `sdl-freerdp3` first, fallback to older logic. Otherwise, use the older logic + `sdl-freerdp3` as fallback.

Hi-Angel: > since we know that sdl-freerdp3 performs badly on some older installations, looking for sdl- version afterwards should minimize the number of issues on older systems.

This is a good point!

Since the performance discussion came from two issues on FreeRDP, I think it may be wise to CC @akallabeth here in case they might have comments regarding, which version is better to detect on Wayland first performance-wise. Feel free to ignore the thread though 😊

akallabeth: @Hi-Angel well, performance is the wrong measure here (basically both do the same, open a window and blit bitmaps to it, so they should not differ much)
but some feedback to sdl-freerdp is the client we're trying to promote as reference (since it works on both, x11 and wayland)

petersilva: @Hi-Angel ok in that case, if sdl is to be *promoted*, then we should put it first?  Fine with me... upto you...

Hi-Angel: > upto you...

Well, I'm not a maintainer, so it's not up to me, let's wait for their comments 😊

waffles-dev: My natural response without knowing _too_ much about this area myself (and not having an opinion on which freerdp is best) is to have an aversion to changing how things currently work for users who currently have a working app.
If this PR fixes the issue in its current state then I'd be happy to take it as-is. But I'll check with Tibix/Levev if they have any concerns first.

petersilva: > My natural response ... is an aversion to changing how things currently work for users who currently have a working app ...

yes: The way the PR is currently structured, it adds a last ditch option that might succeed if all others have failed.  It changes no currently successful code paths.  So it's already the most conservative option.

 The way people are suggesting to change the PR  by putting the sdl-freerdp earlier in the search list would change the freerdp chosen for some people who have working configurations and introduce potential breakage.  so what is currently submitted is the least likely to introduce any breakage at all.

MilitaoLucas: > My natural response without knowing _too_ much about this area myself (and not having an opinion on which freerdp is best) is to have an aversion to changing how things currently work for users who currently have a working app. If this PR fixes the issue in its current state then I'd be happy to take it as-is. But I'll check with Tibix/Levev if they have any concerns first.

Couldn't you add a new config parameter that has a default value of sdl-freerdp/sdl-freerdp3 for wayland systems? That way, if it is not define in the json config, the program assumes the user is a legacy user who just updated. Basically transfer the options for freerdp packages to the user config and if the flag isn't defined add the legacy option, with sdl-freerdp as the last option and if it is defined add it as a first option. 

Because for people who use Wayland the package appears very broken under a lot of stuff like Visual Studio when using Xwayland. 
Also, another detail is that some distros ship sdl-freerdp instead of sdl-freerdp3, so that should be added as well.

petersilva: What do you mean by "config parameter" there is no configuration in this code.   It is code that makes guesses about the freerdp binary available without any configuration.  If you mean adding an additional guess to try freerdp without the 3 at the end, sure I could do that.  I haven't seen any cases where the client is packaged that way... So I'd like to see some evidence of the existence of such a thing... but there's no technical barrier to adding that.

MilitaoLucas: > I haven't seen any cases where the client is packaged that way... So I'd like to see some evidence of the existence of such a thing... but there's no technical barrier to adding that.

NixOS pkgs:
```
❯ ls /run/current-system/sw/bin/ | grep freerdp
freerdp-proxy
freerdp-shadow-cli
sdl-freerdp
sfreerdp
sfreerdp-server
wlfreerdp
xfreerdp
```
> What do you mean by "config parameter" there is no configuration in this code. It is code that makes guesses about the freerdp binary available without any configuration. If you mean adding an additional guess to try freerdp without the 3 at the end, sure I could do that. 

The problem is that this approach will not work. Most people/distros use Xwayland. X11 is still very prevalent. So this will affect almost zero users. If you try first xfreerdp it will work more than 90% of the time. I don't think an automated approach will work for this reason.

petersilva: The reason I submitted a PR was to have maintainers review it and provide feedback, with the hope of getting it accepted, as the current code base requires work-arounds in my system.

There is an infinity of possible approaches, but the actual constraint is what the maintainers are willing to review and accept.  This one line PR is a month old, and as far as I know hasn't been reviewed by an actual maintainer.  That's not a complaint.  I'm hearing a lot about projects being swamped with AI-bot generated PR's, and people I totally get having backlogs and priorities.
Still, grand discussions about replacing a one line patch with a configuration system are pointless if we don't have feedback from the people who support the code.  All we are doing is giving the poor maintainer a huge novel to read, around a 1 line patch.

I don't think random other people commenting is helpful. If it doesn't make the PR any more likely to be approved, it doesn't provide any useful feedback, because I don't know if the maintainers agree with the views or not.
So please, if you are not involved in whatever this project's formal code review process is, then please appreciate that your comments just make more work for everyone involved, and are not really helpful.

Answering the issues raised (again) anyways:

* The *automated approach* is what is currently implemented for all other cases.  Changing the approach changes this PR from one line to hundreds. scope creep on steriods.
* The *automated approach* works for you, since the current code already finds a working client.  Other commenters wanted to make sure that this doesn't break or change any currently working use cases.  The existing PR achieves that.
* People who prefer the sdl-client over the xclient have the option of removing the x client.  On my OS, the clients are all in different packages ( freerdp3-wayland, freerpdp3-sdl, ... there isn't an x one.)
* If there is a real desire to complicate the code base by adding configuration options, then I can certainly do that, but I'd like to hear from an actual maintainer before doing work that is likely to be rejected.
* If there is a real desire to prefer the SDL client over the x one... that is trivially easy to do, but it would perhaps change behaviour on some systems and perhaps generate support load or issues, so again, I would want the maintainer to ask for that.


The behaviour of the current patch is minimal on purpose. I don't know if an actual maintainer has looked at it yet.  I won't change anything unless a maintainer asks.

waffles-dev:
I'm more than happy with this PR in it's current form - I don't think any changes need to be made. I've raised this one with Tibix and he was saying he still has issues on his machine with the SDL version. And I guess he's a bit wary of opening us up to supporting SDL issues that others may have while his is still in such a state.

This isn't to say this is a wasted effort to try and get it, my current assumption is that we'd take it as-is, but it just might take us a little bit of time to getting around to this one - sorry!

src/renderer/utils/getFreeRDP.ts
MilitaoLucas: Fair enough. I was just pointing some stuff up. About the package without the 3 I think it is worth it as it is just another line of code. Sorry for making this harder. I will make another PR if this is merged with my proposed solutions. I wouldn't want to impose on you if you already solved it for your use case.

Self-Hosting-Group:
Fedora used `sdl-freerdp` and not `sdl-freerdp3`

akallabeth:
@Self-Hosting-Group that is dependent on compilation options. 
names can be:
* `sdl-freerdp`
* `sdl-freerdp3`
* `sdl3-freerdp`
* `sdl3-freerdp3`

the scheme is `sdl<sdl version>-freerdp<freerdp version>`

petersilva:
2nd commit added, added additional choices.

Self-Hosting-Group:
> 2nd commit added, added additional choices.

Thanks, with `sdl-freerdp` the client can now also be found under Fedora. However, without deleting the `xfreerdp` executable, it cannot be currently given preference (e.g. with Wayland), as the same package contains both clients.

For the `sdl3-*` variants. Did you do this for a specific distribution? Since I couldn't find any that use this name after a quick search, I personally wouldn't add these names.

Hi-Angel:
> For the `sdl3-*` variants. Did you do this for a specific distribution? Since I couldn't find any that use this name after a quick search, I personally wouldn't add these names.

As noted above, these can appear depending on the compilation options. Some distro may change options they use making such name appear.

akallabeth:
@Self-Hosting-Group this is controllable by build options. (way too complicated, but there are just too many different use cases)

* if some distro wanted to ship both, `SDL2` and `SDL3` versions it is activated automatically.
* the default name is `sdl-freerdp` (with the optional 3 which allows side by side installation with freerdp 2), but that can be both, `SDL2` or `SDL3` (depending on which version was compiled)

that being said, `FreeRDP 2` is now ending its lifecycle and `SDL3` is the preferred choice, the legacy stuff should not matter much anymore.

Self-Hosting-Group:
> the default name is `sdl-freerdp` (with the optional 3 which allows side by side installation with freerdp 2), but that can be both, `SDL2` or `SDL3` (depending on which version was compiled)

Note that distributions with newer codebases (e.g. Fedora) no longer include these earlier freerdp/SDL versions.

> As noted above, these can appear depending on the compilation options. Some distro may change options they use making such name appear.

Personally, I would simply not insert any executeable names that do not currently exist, in order to avoid bloating the code and for security reasons.

But it doesn't bother me either; I'm happy with the way things are at the moment.

Twig6943:
Any updates on this? Gnome deprecated x11 with gnome 50 since the last activity on this

petersilva:
afaict no actual dev has even looked at it.  no feedback received.  Thanks to the others who have commented, but this doesn't seem any closer to getting pulled.

waffles-dev:
> afaict no actual dev has even looked at it. no feedback received. Thanks to the others who have commented, but this doesn't seem any closer to getting pulled.

I looked at it and gave feedback: https://github.com/TibixDev/winboat/pull/649#issuecomment-3912164436

Where it goes from here, I'm not sure just yet. It's looking like next release is not likely to be very soon, but I'll try to get a final decision/answer for you before that time.

petersilva:
> > afaict no actual dev has even looked at it. no feedback received. Thanks to the others who have commented, but this doesn't seem any closer to getting pulled.
> 
> I looked at it and gave feedback: [#649 (comment)](https://github.com/TibixDev/winboat/pull/649#issuecomment-3912164436)
> 
oops.  I guess I didn´t know... I guess it´s hard for me to tell, because without a formal group, I don´t know who is a real contributor and who is not, because it is in an individual developer´s repo.  Thanks for your efforts and explanations.

Hi-Angel:
Maintainers usually have a badge saying something different from "contributor" and "reporter". I think, "collaborator", "owner" and "maintainer", unless I'm confusing some terms with Gitlab

waffles-dev:
> Maintainers usually have a badge saying something different from "contributor" and "reporter". I think, "collaborator", "owner" and "maintainer", unless I'm confusing some terms with Gitlab

For whatever reason it's Tibix as owner, then Levev and myself as Contributors, then everyone else.
But the Maintainer role is there on discord.

Hi-Angel:
> then Levev and myself as Contributors

You are shown as Collaborator

image

😊

waffles-dev:
> > then Levev and myself as Contributors
> 
> You are shown as Collaborator
> image
> 
> 😊

I mistyped. Point still stands.

petersilva:
#646 has been closed as "not planned" ... the claim is that the the sdl implementation "doesn't support remoteapps" ... tried looking up remote apps in teh freerdp project... to no avail. I don't know what is meant.  It seems to work fine on my laptop, I don't know what the breakage is supposed to be.  Anyways, if the bug is "Not Planned", then the PR doesn't make any sense.

terraquad:
The client not supporting remoteapps is `wlfreerdp`, `sdl-freerdp` is a different client which seems to work (it's not mentioned in the referenced docs but on wayland it works much better than xfreerdp [better window placement]). so this pull request is still valid imo...

BinToss:
> The client not supporting remoteapps is `wlfreerdp`, `sdl-freerdp` is a different client which seems to work (it's not mentioned in the referenced docs but on wayland it works much better than xfreerdp [better window placement]). so this pull request is still valid imo...

It doesn't. See https://github.com/FreeRDP/FreeRDP/issues/9078 , https://github.com/FreeRDP/FreeRDP/pull/13133

PR #660, opened 2026-01-29
feat: Configuration screen code cleanup

Levev:
Separating out the input cards into their own component already makes a night and day difference.
It massively reduces development debt, making it much easier to work with.

With this we lay the foundation for a Configuration screen overhaul, as it's already quite all over the place


TibixDev:
This is beautiful, much-much cleaner than it was and this way we got rid of some rerender pesky hacks too. LGTM! 🙌

PR #675, opened 2026-02-05
feat: Break config up into separate sections

Levev:
The config screen was getting way too cluttered for its own good, so breaking it up into separate sections will definitely improve UX.
The changes included in this PR at the time of its creation are very much conceptual, and will most likely change significantly later down the line.
Right now the biggest change is how routing is handled imo, as case information is now included in the path itself, and routes define both the icon that'll be used in the navbar (if the nav property is true) and config screen.
The config buttons are very wonky right now haha.


Levev:
Should be ready for review

TibixDev (changes requested):
Overall this is a nice overhaul concept, I definitely agree that our Config view is currently extremely cluttered up.

There's some things that I'd like to request even before we get to deciding how the categories should exactly work:
- The SVG icon (`icons/winboat_iconify.svg`) should reside either in `assets` (preferred if possible) or `public`
- `Configuration` here should become gray (and all of these should be `select-none` perhaps) once we go into a subroute, and it should be clickable to go back to the base Configuration screen
image<br>
- Configuration subroutes should also show up here, indented somehow to look visually pleasing, they should also be clickable for the user, with the same effect as selecting a category<br> image
- While I like the unsquish animation when entering a Configuration subroute, shouldn't it start from a later point (e.g. 70%) to look less janky? Otherwise maybe we can keep our original fade animation for this too, whatever looks nicer.
- The route meta for Configuration subroutes should also include a description entry I assume
- Don't inline for `v-for` for `ConfigButton` since it's rather long
- Shouldn't hover an option when my mouse is there<br> image

eriperdana598-netizen (approved):

PR #705, opened 2026-03-03
Readme improve for podman installation

Miit0s:
When I try for the first time to create a podman with WinBoat, the podman creation failed with an error like this (I have deleted the log files, so it is not hyper precise sorry) :  `"podman compose -f" command failed, flag "f" doesn't exist"`. After a lot of debugging, I simply realised that my podman version was too old (because the one installed with apt install on Pop_OS! is old), and this was causing the error.
So I propose this change in the README to avoid other users like me loose time for something that is simple.

I'm not quite sure of the format; it's not very similar to what we have at the moment, but feel free to let me know where to make changes or do the change by yourself so that it fits better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,7 @@ Before running WinBoat, ensure your system meets the following requirements:
 - **In case of Podman:**
   - **Podman**: Required for containerization
       - [Installation Guide](https://podman.io/docs/installation#installing-on-linux)
-
-      > On Debian/Ubuntu and forks, the Podman version installed with `apt install` can be too old. Make sure you have **Version 4.x.x** or higher to have the installation complete successfully
+      > On Debian/Ubuntu and forks, the Podman version installed with `apt install` could be too old. Make sure you have **Version 4.x.x** or higher to ensure the installation completes successfully.
   - **Podman Compose**: Required for compatibility with podman-compose.yml files
       - [Installation Guide](https://github.com/containers/podman-compose?tab=readme-ov-file#installation)
 - **FreeRDP**: Required for remote desktop connection (Please make sure you have **Version 3.x.x** with sound support included)
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Before running WinBoat, ensure your system meets the following requirements:
 - **In case of Podman:**
   - **Podman**: Required for containerization
       - [Installation Guide](https://podman.io/docs/installation#installing-on-linux)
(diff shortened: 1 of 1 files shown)


LittleT6109 (changes requested):
Multiple issues. 
- "fork" should be "forks"
- Capitalize "podman"
- "install" should be "installed"
- Move the note into the list on the same level as the Podman installation guide

README.md
```diff
@@ -58,6 +58,8 @@ Before running WinBoat, ensure your system meets the following requirements:
 - **In case of Podman:**
   - **Podman**: Required for containerization
       - [Installation Guide](https://podman.io/docs/installation#installing-on-linux)
+
+      > On Debian/Ubuntu and forks, the Podman version installed with `apt install` can be too old. Make sure you have **Version 4.x.x** or higher to have the installation complete successfully
```
LittleT6109:
```suggestion
      - On Debian/Ubuntu and forks, the Podman version installed with `apt install` could be too old. Make sure you have **Version 4.x.x** or higher to ensure the installation completes successfully.
```
Sorry, still new to reviewing and didn't realize I could do this.

LittleT6109 (changes requested):

README.md
Miit0s: Do you thinks it's better to remove the "note style" display by switching > to - ?

LittleT6109: yes, that way it is the same style as the docker desktop note.

README.md
```diff
@@ -58,8 +58,7 @@ Before running WinBoat, ensure your system meets the following requirements:
 - **In case of Podman:**
   - **Podman**: Required for containerization
       - [Installation Guide](https://podman.io/docs/installation#installing-on-linux)
-
-      > On Debian/Ubuntu and forks, the Podman version installed with `apt install` can be too old. Make sure you have **Version 4.x.x** or higher to have the installation complete successfully
+      > On Debian/Ubuntu and forks, the Podman version installed with `apt install` could be too old. Make sure you have **Version 4.x.x** or higher to ensure the installation completes successfully.
```
LittleT6109:
```suggestion
      - **⚠️ NOTE:** On Debian/Ubuntu and forks, the Podman version installed with `apt install` could be too old. Make sure you have **Version 4.x.x** or higher to ensure the installation completes successfully.
```

LittleT6109:
forgot the **⚠️ NOTE:** but good enough i guess

LittleT6109 (approved):
missing the **⚠️ NOTE:** but otherwise LGTM

waffles-dev (approved):

Issue #557, opened 2025-12-07
[Bug] novnc


rreyn331:
### What version of WinBoat are you running?

.90

### Your Environment

Distro: Debian Forky
Winboat versio 9.0
Desktop Environment XFCE 4.2
FreeRDP3-x11 3.18.0+dfsg1

### Steps to Reproduce / Context

upgrade from winboat .87 to .90 and no l0nger able to access docker containerthrough website at http://127.0.0.1:8006/

### Logs

[containera.log](https://github.com/user-attachments/files/24015826/containera.log)

[docker-compose-crasha.txt](https://github.com/user-attachments/files/24015827/docker-compose-crasha.txt)

[docker-compose-worka.txt](https://github.com/user-attachments/files/24015829/docker-compose-worka.txt)

[winboat.log](https://github.com/user-attachments/files/24015838/winboat.log)

### Expected Behavior

open up browser to http://127.0.0.1:8006/ and have access to log into docker windows

### Current Behavior

http://127.0.0.1:8006/ opens up with unable to connect

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

rreyn331:
The issue has been solved web site to access changed port numbers so it is now http://127.0.0.1:47271/ instead of http://127.0.0.1:8006 just wish it was documented elsewhere

waffles-dev:
For podman, it will be 8096 by default, but we were having port collisions, so instead we just go to an available port. The UI should present the correct url/port for VNC connections, and I'm unable to reproduce on latest main.

PR #708, opened 2026-03-04
fix novnc issue

NextWork123:
Hello,

I saw issue #557 and was investigating the problem. I deployed WinBoat using Podman and noticed that it assigns a random port on the host side. However, we need port 8006 to be exposed on the host as well.

Best regards,
NextWorks

### DIFF
--- a/src/renderer/data/podman.ts
+++ b/src/renderer/data/podman.ts
@@ -26,7 +26,7 @@ export const PODMAN_DEFAULT_COMPOSE: ComposeConfig = {
             },
             cap_add: ["NET_ADMIN"],
             ports: [
-                "127.0.0.1::8006", // VNC Web Interface
+                "127.0.0.1:8006:8006", // VNC Web Interface
                 "127.0.0.1::7148", // Winboat Guest Server API
                 "127.0.0.1::7149", // QEMU QMP Port
                 "127.0.0.1::3389/tcp", // RDP


waffles-dev:
Did yours really need this to work?
Mine works with a podman compose of

```
    ports:
      - "127.0.0.1::8006"
      - "127.0.0.1::7148"
      - "127.0.0.1::7149"
      - "127.0.0.1::3389/tcp"
      - "127.0.0.1::3389/udp"
```

I'd like to know what the root cause was that meant you needed that change. Not saying this PR doesn't have merit, but I'd like to understand what's going on first.

NextWork123:
> Did yours really need this to work? Mine works with a podman compose of
> 
> ```
>     ports:
>       - "127.0.0.1::8006"
>       - "127.0.0.1::7148"
>       - "127.0.0.1::7149"
>       - "127.0.0.1::3389/tcp"
>       - "127.0.0.1::3389/udp"
> ```
> 
> I'd like to know what the root cause was that meant you needed that change. Not saying this PR doesn't have merit, but I'd like to understand what's going on first.

Hello,
i use the automate installer of winboat (with the selection of podman) and it wasn't working as default so the PR is needed now i don't have the recent updates or so other because after that i didn't use anymore.

waffles-dev:
Ahk, thanks for the update. We'll leave it out for now, the goal is to avoid port collisions, and latest main seems to find the actual port used for vnc properly to support connections.
Thanks though!

PR #714, opened 2026-03-06
feat: Add system tray icon

LittleT6109:
Inspired by [this](https://discord.com/channels/1408451891523752017/1428850296452284636) tidbit's idea. I did not download the tidbit as it seemed to have some issues based on the forum history.

https://github.com/user-attachments/assets/840db0b5-5b8e-442b-9bca-11b6a2b87bfa


softwaresocialist:
That would be great

PR #735, opened 2026-03-27
Add Linux Arm Support

easton57:
Hey there,

This uses journalctl to check for EL2 and virtualization support. Unfortunately arm doesn't have virtualization features listed in /proc/cpuinfo.


gitneko:
You should wrap the journalctl check into its own `try` block, because on non-ARM devices this will completely fail the virtualization check (due to a non-zero exit code of `grep` it will go to the `catch` block).

easton57:
I'll work on adding it to a try block in the coming week.

easton57:
Sorry had some life events that restricted my time. I'll take a look this week for sure if support hasn't already been added.

easton57:
Closing as I accidentally committed other changes that were personal to the same branch.

PR #755, opened 2026-04-18
Fix: quote strings in podman compose

waffles-dev:
Some users have reported that the restart-always/never flag in podman compose yml doesn't work unless they have quotes in the compose. I believe this is a compatibility issue with the podman-compose binary.
This will also protect from special characters in username, password and file path fields.

By flagging defaultStringType as quote double, all default types in the stringify will be quoted.
defaultKeyType as plain then overrides the quotes for the keys in the yml, leaving them unquoted as before.


TibixDev:
Looks good to me, thank you! 🙌

PR #757, opened 2026-04-18
Fix stale prerequisite validation when switching from Docker to Podman

### LINKED ISSUE #756: [Bug] You can still continue switching from Docker to Podman even if the conditions are not met.

0.90

### Your Environment

Ubuntu24.04LTS

There's a bug here: if all conditions are met (docker), when switching from the Docker tab to the Podman tab, the user can still click Next to continue, instead of waiting for the conditions to be met.
docker(normal):

Image
Podman(Bug):

Image

### Logs

nome

### Expected Behavior

This should not continue; we should wait until all conditions are met.

### Current Behavior

Under the current circumstances, users can continue even if the conditions are not met.

### Possible Solution

_No response_

### Quality Notice

Iamliuxiaozhen:
## Summary

This fixes a setup wizard bug where the `Next` button could remain enabled after switching the container runtime from Docker to Podman, even though Podman prerequisites had not been validated yet.

## What changed

- replaced the async computed `containerSpecs` state with an explicit `watch` on `containerRuntime`
- clear the previous runtime's prerequisite state immediately when the runtime changes
- added a request id guard so outdated async checks cannot overwrite the latest runtime state

## Why

The previous implementation could briefly reuse the old runtime's successful prerequisite result while the new runtime check was still in flight. That allowed users to continue with stale validation state.

## Testing

- switched between Docker and Podman in the setup prerequisites step
- confirmed the `Next` button is disabled immediately after switching runtimes
- confirmed the button only becomes enabled again after the current runtime's prerequisites pass

## Related

- Closes #756

### DIFF
--- a/src/renderer/views/SetupUI.vue
+++ b/src/renderer/views/SetupUI.vue
@@ -810,6 +810,7 @@
 
 <script setup lang="ts">
 import { Icon } from "@iconify/vue";
+import { computedAsync } from "@vueuse/core";
 import { computed, onMounted, onUnmounted, ref, watch } from "vue";
 import { useRouter } from "vue-router";
 import { InstallConfiguration, Specs } from "../../types";
--- a/src/renderer/views/SetupUI.vue
+++ b/src/renderer/views/SetupUI.vue
@@ -812,7 +812,6 @@
 import { Icon } from "@iconify/vue";
 import { computed, onMounted, onUnmounted, ref, watch } from "vue";
 import { useRouter } from "vue-router";
-import { computedAsync } from "@vueuse/core";
 import { InstallConfiguration, Specs } from "../../types";
 import { getSpecs, getMemoryInfo, defaultSpecs, satisfiesPrequisites, type MemoryInfo } from "../lib/specs";
 import { WINDOWS_VERSIONS, WINDOWS_LANGUAGES, type WindowsVersionKey } from "../lib/constants";
@@ -821,6 +820,7 @@ import { openAnchorLink } from "../utils/openLink";
 import license from "../assets/LICENSE.txt?raw";
 import {
     ContainerRuntimes,
+    type ContainerSpecs,
     DockerSpecs,
     PodmanSpecs,
     getContainerSpecs,
@@ -936,6 +936,7 @@ const sharedFolderPath = ref("");
 const installState = ref<InstallStates>(InstallStates.IDLE);
 const preinstallMsg = ref("");
 const containerRuntime = ref(ContainerRuntimes.DOCKER);
+let containerSpecsRequestId = 0;
 const vncPort = ref(8006);
 // These are the install steps where the container is actually up and running
 const linkableInstallSteps = [ InstallStates.MONITORING_PREINSTALL, InstallStates.INSTALLING_WINDOWS, InstallStates.COMPLETED ];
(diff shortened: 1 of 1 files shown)


src/renderer/views/SetupUI.vue
```diff
@@ -812,7 +812,6 @@
 import { Icon } from "@iconify/vue";
 import { computed, onMounted, onUnmounted, ref, watch } from "vue";
 import { useRouter } from "vue-router";
-import { computedAsync } from "@vueuse/core";
```
waffles-dev, line 815: Why was this removed? Still seeing it used lines 1084 and 1116

Iamliuxiaozhen:
Good catch, that was my mistake. I removed the import while replacing the async prerequisite check for `containerSpecs`, but `computedAsync` is still used below for the install folder validations. I’ve added the import back in 8023b98.

waffles-dev:
> Good catch, that was my mistake. I removed the import while replacing the async prerequisite check for `containerSpecs`, but `computedAsync` is still used below for the install folder validations. I’ve added the import back in [8023b98](https://github.com/TibixDev/winboat/commit/8023b9843096fc77cec2e7189be2b7b946196a11).

Please restore it to the original line so there's no diff, thanks :)

TibixDev (changes requested):
Assuming `computedAsync` is what breaks here for some reason, wouldn't it be easier to have a simple watch that directly reassigns `containerSpecs`? The operation isn't heavy enough to require the duplication check (or if it is, let me know), so we can avoid having a request counter and all that stuff.

Shouldn't it be like two lines that way? Just reassign as `ref` as you did and use an immediate watcher on the runtime.

Let me know. 😄

TibixDev:
Oh and additionally, if it does take a longer time, we can have some boolean ref such as `checkingPrerequisites` that'd turn false/true before and after the check, and conditionally disable the button with `<existing> || checkingPrerequisites`

src/renderer/views/SetupUI.vue
```diff
@@ -972,9 +974,23 @@ watch(folderSharing, (newValue) => {
...
 
-const containerSpecs = computedAsync(async () => {
-    return await getContainerSpecs(containerRuntime.value);
-});
+const containerSpecs = ref<ContainerSpecs>();
+
+watch(
+    containerRuntime,
+    async runtime => {
+        const requestId = ++containerSpecsRequestId;
```
waffles-dev: this doesn't actually fix the issue.
I can open the drop down, tab to podman (which I've set to fail pre-reqs), then press enter and immediately click the Next button

I think a better approach is to immediately disable the Next button when changing selections in the picker. Then have a watcher on containerSpecs that re-enables the Next button when they're passing the checks.

Iamliuxiaozhen:
Thanks, updated this to immediately mark prerequisites as checking from the runtime picker change handler, clear the previous specs synchronously, and disable Next while the current runtime checks are running. I also simplified the watcher and removed the request counter.

waffles-dev:
> Thanks, updated this to immediately mark prerequisites as checking from the runtime picker change handler, clear the previous specs synchronously, and disable Next while the current runtime checks are running. I also simplified the watcher and removed the request counter.

I can still click fast enough after selecting Podman in the drop down that it lets me through when it shouldn't.

Iamliuxiaozhen:
Thanks for re-testing. I moved the guard earlier than the `x-select` change event by marking prerequisites as checking from the runtime menu item interaction itself (`pointerdown`, Enter, and click). I also changed the Next button click handler to re-check the prerequisite state before advancing, so even if the disabled state has not visually updated yet, the click path should no longer let the wizard through.

waffles-dev:
> Thanks for re-testing. I moved the guard earlier than the `x-select` change event by marking prerequisites as checking from the runtime menu item interaction itself (`pointerdown`, Enter, and click). I also changed the Next button click handler to re-check the prerequisite state before advancing, so even if the disabled state has not visually updated yet, the click path should no longer let the wizard through.

Thanks. It's working well for me. Bear with us - we'll get it reviewed eventually :)

Iamliuxiaozhen:
Friendly ping 👋

Just checking in on this PR. Since the last update was a while ago, I wanted to see if there's anything else needed from my side.

Thanks!

waffles-dev:
> Friendly ping 👋
> 
> Just checking in on this PR. Since the last update was a while ago, I wanted to see if there's anything else needed from my side.
> 
> Thanks!

Hey, sorry, we're thinking this is probably a bit overcomplicated for what it needs to do.
Have you seen Tibix's comments above?

PR #768, opened 2026-04-30
chore: bump minor depends and fix issues

a5ehren:
Bumped all depends to latest minor version, which exposed these issues that I fixed:

- jimp did not export itself the same way so build failed. replaced with Canvas operations and removed jimp
- tailwindCSS warns about "false" not being a valid value for darkMode so changed to "media"
- build.ts failed silently if either compiler failed. Just nuked it and put it in package.json
- added format and type-check scripts to package.json


a5ehren:
closing because I can't test it

PR #770, opened 2026-05-02
flatpak builders

guidryheal-create:
the git action still need to be tested but no reason not to workout, local build, launch and etc seems to work

you can still do extended test with the realised .flatpak

should help you debug and overall sandbox the sandbox 

flathub is awesome for both security and convinience so why not use it hehe


guidryheal-create:
any news ? i wonder your interest in that matter

azenyr:
Would love winboat in flathub for sure, auto updates through the distro flatpak software store is awesome for ease of use

a5ehren:
I think #629 does this better, or at least in a more idiomatic way for packaging electron apps in flatpak.

PR #776, opened 2026-05-10
feat: enable support for memory ballooning feature provided by dockurr/windows image

### LINKED ISSUE #522: [Feature] Dynamically resize RAM
### Feature Description

Currently the RAM the VM has seems to be fixed. But some apps may require a lot of RAM temporarily (esp. Adobe Apps). It would be nice if the VM could dynamically request more RAM from the host for the app until maybe a configured maximum. After the app closes it would give back the RAM to the host and return to the default assigned RAM.

### Use-Case

Apps which temporarily need a lot of RAM like some Adobe Apps.

### Proposed Implementation

Not sure if this is possible but I think KVM can do this in theory.

### Notice

- [x] I would like to help out implementing this feature

lukakama:
### Summary
Adds support for dynamic memory ballooning in Windows VMs, allowing the host system to reclaim unused memory when host is under memory pressure. This feature integrates with the new dockurr/windows image's built-in ballooning feature released on 5.15 (https://github.com/dockur/windows/releases/tag/v5.15).

Fix: #522

**Main changes:**
- New experimental feature configuration to enable or disable dynamic memory (ballooning)
- Setup of the Windows ballooning service for already installed Windows containers; newly installed Windows containers get the service installed from dockurr/windows unattended installation
- Report of current effective VM available and used memory with balloon-corrected used memory
- Use of QEMU QMP commands correlation ID, due to possible concurrent commands execution, with timeout support

#### Guest Server
- Added `/balloon/status` endpoint to check balloon service status
- Added `/balloon/install` endpoint to install the balloon service

#### QMP Manager
- Added `query-balloon` command support
- Refactored QMP connection handling with EventEmitter pattern
- Implemented proper message buffering and command correlation with IDs
- Added configurable timeouts for commands
- Improved error handling and connection lifecycle management

#### Winboat Core
- Integrated balloon status checking via QMP `query-balloon` command
- Auto-installs missing balloon service when experimental features are enabled
- Used dedicated memory stats object to reflect available, used and percentages memory stats accounting for ballooning

#### UI Updates
- Config View
  - Added "Dynamic memory" toggle (visible when experimental features enabled)
  - Persists `BALLOONING=Y` environment variable in docker-compose when "Dynamic memory" is enabled
  - Disables ballooning when experimental features are turned off

- Home View
  - Updated RAM display to show ballooned vs total memory
  - Shows adjusted RAM percentage based on actual available memory
  - Conditionally displays balloon info when memory is reclaimed

Feedback and suggestions for improvement are welcome, particularly for ballooned memory stats reporting.

### DIFF
--- a/guest_server/main.go
+++ b/guest_server/main.go
@@ -127,6 +127,58 @@ func getMetrics(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(metrics)
 }
 
+func getBalloonStatus(w http.ResponseWriter, r *http.Request) {
+	cmd := exec.Command("C:\\Windows\\Drivers\\Balloon\\blnsvr.exe", "status")
+	output, err := cmd.CombinedOutput()
+
+	response := map[string]string{}
+
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			switch exitErr.ExitCode() {
+			case 1060:
+				response["status"] = "not-installed"
+				response["details"] = string(output)
+			case 1, 1062:
+				response["status"] = "stopped"
+				response["details"] = string(output)
+			default:
+				response["status"] = "error"
+				response["details"] = string(output)
+			}
+		} else {
+			response["status"] = "error"
+			response["details"] = err.Error()
+		}
+	} else {
+		response["status"] = "running"
+		response["details"] = string(output)
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(response)
+}
+
+func installBalloon(w http.ResponseWriter, r *http.Request) {
+	cmd := exec.Command("C:\\Windows\\Drivers\\Balloon\\blnsvr.exe", "-i")
+	output, err := cmd.CombinedOutput()
+
+	response := map[string]string{}
+
+	if err != nil {
+		response["result"] = "error"
+		response["details"] = string(output)
+	} else {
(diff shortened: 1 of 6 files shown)


lukakama:
The ballooning feature has just been released on dockur/windows v5.15 (https://github.com/dockur/windows/releases/tag/v5.15).

TibixDev:
Hello, sorry for the late response on your PR. It's been some months since the ballooning feature has released in `dockur/windows`, and if @kroese and you deem it stable, perhaps it doesn't need to be an experimental feature with a toggle, can be enabled by default implicitly, and that way we avoid having to write tons of extra code.

The migration path would be needed, however I do not plan 1.0 at this moment to be backwards compatible, if that changes I can refer to the current state of the PR, but otherwise all of that can be dropped too at this moment.

Waiting on your feedback :)

kroese:
@TibixDev The problem is that it requires a driver to be installed in the Windows guest. Only recent releases of the image now include this driver in the unattended setup, but for a lot of people (who have older installations) the ballooning will not work unless they manually install this VirtIO driver themselves.

So maybe your Guest Agent / API program, can check after boot if the "Ballooning Service" is installed, and if not, automatically install that service. Or otherwise detect the WinBoat version that was used for the  install (or the creation date of the files), and just disable the feature for older installs.

**EDIT:** I now see that this pull request already fixes this problem, never mind.

But it is a really advanced/complicated feature, that many people will not fully understand (including me). And that was not tested a lot. So I would hide it under an Advanced/Experimental option for now, otherwise it will raise a lot of questions.

PR #777, opened 2026-05-12
feat(dev): Add oxlint and a CI workflow for fmt & lint

a5ehren:
The project did not have any mechanism to enforce format settings or code quality, so this PR does:

- Update prettier config to be aware of tailwind CSS
- Add oxlint with a basic config
- Fix the issues that fall out of prettier and oxlint against current HEAD
- Add a CI workflow to check both on all PRs

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -1,35 +1,32 @@
 <template>
-    <main 
-        class="overflow-hidden relative w-screen h-screen"
-        :class="{ animationsDisabled: 'disable-animations' }"
-    >
+    <main class="relative h-screen w-screen overflow-hidden" :class="{ animationsDisabled: 'disable-animations' }">
         <!-- Decoration -->
         <div
-            class="gradient-ball absolute -z-10 left-0 bottom-0 translate-x-[-50%] translate-y-[50%] w-[90vw] aspect-square opacity-15"
+            class="gradient-ball absolute bottom-0 left-0 -z-10 aspect-square w-[90vw] translate-x-[-50%] translate-y-[50%] opacity-15"
         ></div>
         <div
-            class="gradient-ball absolute -z-10 right-0 top-0 translate-x-[50%] translate-y-[-50%] w-[90vw] aspect-square opacity-15"
+            class="gradient-ball absolute right-0 top-0 -z-10 aspect-square w-[90vw] translate-x-[50%] translate-y-[-50%] opacity-15"
             style="filter: hue-rotate(45deg) blur(200px)"
         ></div>
 
         <!-- Stripes for experimental -->
         <div
             v-show="wbConfig?.config.experimentalFeatures"
-            class="experimental-stripes absolute top-0 left-0 w-full h-[3rem] pointer-events-none z-[10] opacity-15 grayscale"
+            class="experimental-stripes pointer-events-none absolute left-0 top-0 z-[10] h-[3rem] w-full opacity-15 grayscale"
         ></div>
 
         <!-- Titlebar -->
         <x-titlebar
             @minimize="handleMinimize()"
             @buttonclick="handleTitleBarEvent"
-            class="backdrop-blur-xl bg-neutral-900/50"
+            class="bg-neutral-900/50 backdrop-blur-xl"
         >
             <x-label>WinBoat</x-label>
         </x-titlebar>
 
         <!-- Updater -->
         <dialog ref="updateDialog">
-            <Icon class="text-indigo-400 size-12" icon="mdi:cloud-upload"></Icon>
+            <Icon class="size-12 text-indigo-400" icon="mdi:cloud-upload"></Icon>
             <template v-if="manualUpdateRequired">
                 <h3 class="mt-2">Manual Guest Server Update Required</h3>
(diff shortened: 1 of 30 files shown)


waffles-dev (commented):
I'm happy with changes to this one. Leaving final call to @TibixDev - he'll need to set up the branch rule anyway.

src/renderer/lib/winboat.ts
```diff
@@ -623,7 +623,7 @@ export class Winboat {
             logger.info(`Found custom app command for '${app.Name}'`);
             customAppCallbacks[app.Path]!(this);
             this.appMgr?.incrementAppUsage(app);
-            this.appMgr?.writeToDisk();
+            await this.appMgr?.writeToDisk();
```
waffles-dev: (sorry for the absence)
actually, this is an overzealous await here, acc. to Tibix we're intentionally ignoring this call

waffles-dev (changes requested):

src/renderer/lib/winboat.ts
a5ehren: cool, linter is happy if we just tag it with `void`

src/renderer/lib/winboat.ts
```diff
@@ -670,7 +670,7 @@ export class Winboat {
         args = args.filter((v, _i, _a) => v.trim() !== "");
 
         this.appMgr?.incrementAppUsage(app);
-        this.appMgr?.writeToDisk();
+        await this.appMgr?.writeToDisk();
```
waffles-dev:
```suggestion
        void this.appMgr?.writeToDisk();
```

waffles-dev (changes requested):
missed one, sry

waffles-dev (commented):
I'm personally happy with changes, I think this helps to improve quality of the project for long term and that this is good to go as-is.
Waiting on feedback from Tibix - he'll have final say on this one.

src/renderer/App.vue
```diff
@@ -208,7 +205,11 @@ onMounted(async () => {
         () => winboat?.isUpdatingGuestServer.value,
         isUpdating => {
             if (isUpdating === true) {
-                novncURL.value = `http://127.0.0.1:${getActiveHostPort(winboat?.containerMgr!, CommonPorts.NOVNC)}`;
```
waffles-dev: Can you revert this one? If the `!` fails we'd rather have the crash and fix a bug that could cause it to be null, rather than silently return.

waffles-dev: (Also, for now, revert any similar instances, if any, of this change in other files)

src/main/main.ts
```diff
@@ -100,13 +100,13 @@ function createWindow() {
...
-        mainWindow.loadURL(`http://localhost:${rendererPort}`);
+        void mainWindow.loadURL(`http://localhost:${rendererPort}`);
     } else {
-        mainWindow.loadFile(join(app.getAppPath(), "renderer", "index.html"));
+        void mainWindow.loadFile(join(app.getAppPath(), "renderer", "index.html"));
     }
 }
 
-app.whenReady().then(() => {
+void app.whenReady().then(() => {
```
waffles-dev: Tibix has asked if we can disable whatever rule is enforcing the explicit void and revert the relevant changes.

src/renderer/tsconfig.json
```diff
@@ -10,7 +10,7 @@
         "resolveJsonModule": true,
         "esModuleInterop": true,
         "lib": ["esnext", "dom"],
-        "types": ["vite/client"]
+        "types": ["vite/client", "web-bluetooth", "w3c-web-usb"]
```
waffles-dev: what were these added for?

.gitignore
waffles-dev, line 1: revert changes to this file for now - this'd need to be considered in a separate PR

src/renderer/lib/winboat.ts
```diff
@@ -121,9 +121,9 @@ class AppManager {
 
         if (this.appCache.values.length == newApps.length && !options.forceRead) return;
 
-        for (const appIdx in newApps) {
-            newApps[appIdx].Usage = this.appCache.find(app => app.Name == newApps[appIdx].Name)?.Usage || 0;
-            this.appUsageCache[newApps[appIdx].Name] = newApps[appIdx].Usage;
+        for (const app of newApps) {
+            app.Usage = this.appCache.find(cachedApp => cachedApp.Name == app.Name)?.Usage || 0;
+            this.appUsageCache[app.Name] = app.Usage;
```
waffles-dev: can this one also be submitted as a separate PR?
does a rule need to be temporarily disabled to allow the old loop to stay in place? - if so, then please do so and enable the rule in the new PR

waffles-dev (changes requested):
Hey, sorry for the long wait, Tibix has finally gotten around to taking a look - have some minor feedback

a5ehren:
Sorry for the noise - rebasing and updating oxlint before addressing the feedback

src/renderer/lib/winboat.ts
a5ehren: done

PR #783, opened 2026-05-23
fix: resolve Vue Ref unwrapping bugs for Guest Server online status

Vikyek:
This PR fixes Vue Ref unwrapping bugs that incorrectly logged and handled offline states in the frontend.

#### Changes

1. **Fix Vue Ref unwrapping bugs ([winboat.ts](file:///home/v/Projects/winboat-git/src/renderer/lib/winboat.ts))**
   - **Issue 1:** The logger checked `this.isOnline` (which is a Vue `Ref` wrapper object) in a ternary expression: `this.isOnline ? "online" : "offline"`. Because the `Ref` wrapper object is always truthy, this always evaluated to `"online"`, which flooded log files with `"went online"` messages even during offline transitions.
   - **Issue 2:** The app launch guard checked `!this.isOnline`, which also incorrectly evaluated to `false`, causing app launches to proceed when the VM was offline and fail silently downstream rather than displaying a clear error.
   - **Fix:** Appended `.value` to both unwrapped `Ref` boolean checks (`this.isOnline.value`).

### DIFF
--- a/src/renderer/lib/winboat.ts
+++ b/src/renderer/lib/winboat.ts
@@ -304,7 +304,7 @@ export class Winboat {
             const _isOnline = await this.getHealth();
             if (_isOnline !== this.isOnline.value) {
                 this.isOnline.value = _isOnline;
-                logger.info(`Winboat Guest API went ${this.isOnline ? "online" : "offline"}`);
+                logger.info(`Winboat Guest API went ${this.isOnline.value ? "online" : "offline"}`);
 
                 if (this.isOnline.value) {
                     await this.checkVersionAndUpdateGuestServer();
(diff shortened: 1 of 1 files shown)


kroese:
@Vikyek The first issue should already be fixed since the [dockur/windows](https://github.com/dockur/windows) image has been updated to v5.15. If your Podman+Passt issue still exists in v5.15, please make an issue in that repo.

waffles-dev:
My view is that of kroese's. Happy to merge the 2nd fix. If the podman change is removed I can merge as-is.

Not saying we won't make the change, if you feel strongly that it's still needed post the 5.1.5 uprev, then raise another PR and we can discuss that one there. Tibix or Levev would likely need to look into that one.

waffles-dev:
Thanks @Vikyek

PR #791, opened 2026-06-06
Update Windows Docker image version to v5.16

kroese:
Update the Docker image to v5.16. There are no changes that can affect WinBoat, so it should be fully backwards compatible.


waffles-dev:
Thanks @kroese!

Issue #745, opened 2026-04-06
AppImage relies on deprecated FUSE2


x9cf9w73cv7y:
### Feature Description

The WinBoat AppImage that is currently available depends on the FUSE2 interface, which is both deprecated and unmaintained. This causes security issues because it forces users to install fuse2. In addition, some modern distributions, such as Fedora 44, no longer support fuse2.

It's possible to avoid this by building the appimage with the type2 runtime
https://github.com/AppImage/type2-runtime.

### Use-Case

Using it on any Distribution that don't include fuse2

### Notice

- [x] I have checked the issue tracker and verified that this Feature Request has not been proposed before.
- [ ] I would like to help out implementing this feature

ABN-Speed:
This change has made it to Fedora 44 which was released on 4/28/26. I have upgraded my Fedora Kinoite to 44 and can no longer run the AppImage. 

```
user@kinoite/var/home/user/AppImages$ ./winboat-0.9.0-x86_64.AppImage 
fuse: failed to exec fusermount: No such file or directory.

AppImages require FUSE to run. 
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
```

sikraemer:
Same Here (Fedora Kinoite 44)

PR #793, opened 2026-06-08
Chore: Uprev appimage to drop FUSE2

waffles-dev:
Uprevving electron-builder to latest version, which supports setting toolset.appimage runtime to 1.0.3.
This should allow us to move off the FUSE2 version of appimage, which satisfies #745.


waffles-dev:
@TibixDev I'm unable to build locally as I can't compile libusb, but I can get past the electron-builder step that cares about this flag. Are you able to run a build on this branch and verify the appimage launches ok?

waffles-dev:
Update: Can build locally. But appimage doesn't run properly, resources not being bundled right. Still need to figure things out.

PR #808, opened 2026-07-03
fix: cache check never working, port index 0 bug, config write per key

fix2015:
Three bugs I found:

- \`appCache.values.length\` was always \`undefined\` because \`Array.values()\` returns an iterator which doesn't have a \`.length\` property. The early-return cache check never fired, meaning every \`updateAppCache\` call re-processed the full app list unnecessarily. Changed to \`appCache.length\`.

- \`hasShortPortMapping\` returned \`false\` for a port at index 0 because \`!!0\` is \`false\`. If the first port entry matched, it was invisible. Changed to \`!== undefined\`.

- The config repair loop was writing the config file to disk inside the loop for each missing key, instead of once after all keys are patched. Moved the write outside.

### DIFF
--- a/src/renderer/lib/config.ts
+++ b/src/renderer/lib/config.ts
@@ -175,8 +175,8 @@ export class WinboatConfig {
             console.log("Successfully read the config file");
 
             // Some fields might be missing after an update, so we merge them with the default config
+            let hasMissing = false;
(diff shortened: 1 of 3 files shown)


waffles-dev (approved):
Thanks for these fixes!

Issue #663, opened 2026-01-31
[Bug] Winboat can't restart podman machine on its own after USB configuration changes


CJ-does-code:
### What version of WinBoat are you running?

Beta v.0.9.0 Prod Appimage

### Your Environment

# Machine specs
<img width="1237" height="410" alt="Image" src="https://github.com/user-attachments/assets/67659391-2b30-4255-8b1e-dd192f45746c" />
podman version 5.7.1

### Steps to Reproduce / Context

1. Install a podman machine using winboat and start it up
2. unplug a USB device (In this case, my microphone)
3. Attempt a restart (sleep,shutoff,etc)

### Logs

`2026-01-31 01:22:35 | INFO | Launching app: ⚙ Windows Desktop at path WINDOWS_DESKTOP
2026-01-31 01:22:35 | INFO | Launching app: ⚙ Windows Desktop at path WINDOWS_DESKTOP
2026-01-31 01:22:35 | INFO | Launch FreeRDP with command:
xfreerdp3  /u:cj  /p:Fucker1  /v:127.0.0.1  /port:41185  /cert:ignore  +clipboard  /sound:sys:pulse  /microphone:sys:pulse  /floatbar  /compression  +f  /scale:100
2026-01-31 01:37:14 | INFO | USB device detached: [046d:c093] Logitech, Inc. | M500s Optical Mouse
2026-01-31 01:37:15 | INFO | USB device attached: [046d:c093] Logitech, Inc. | M500s Optical Mouse
2026-01-31 02:19:03 | INFO | FreeRDP disconnected due to user logging off.
2026-01-31 02:19:04 | INFO | Winboat Guest API went online
2026-01-31 02:19:11 | INFO | Winboat Container state changed to Exited
2026-01-31 02:19:11 | INFO | Destroying Winboat API intervals...
2026-01-31 07:01:09 | INFO | Starting WinBoat container...
2026-01-31 07:01:09 | ERROR | There was an error performing the container action.
2026-01-31 07:01:09 | ERROR | Error: Command failed: podman container start WinBoat
Error: unable to start container "ff3b71d6282d8d8ccb951c94583f29053fa68ffc69b0d098c5d0195027f74ef2": crun: cannot stat `/dev/bus/usb/001/004`: No such file or directory: OCI runtime attempted to invoke a command that was not found

2026-01-31 07:06:15 | INFO | Final USB database file path: /usr/share/hwdata/usb.ids
[...]

Issue #591, opened 2025-12-21
[Bug] Can't start Windows VM (Loading Infinitely)


ShaunRoselt:
### What version of WinBoat are you running?

0.9.0 Prod

### Your Environment

<img width="1920" height="1080" alt="My SteamOS computer's fastfetch and kinfo specs. I use this computer for work such as Delphi Programming and other stuff" src="https://github.com/user-attachments/assets/c76014b3-0012-4314-898c-6d951bf49e79" />

### Steps to Reproduce / Context

Just try to start it, you'll see it loads forever and never starts:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/17f45797-3c20-4c8d-b296-125d02540d90" />

### Logs

I have no idea where to get these logs mentioned here.

### Expected Behavior

It should start the Windows VM

### Current Behavior

Loads forever

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

ShaunRoselt:
Ah. I think the issue is because Podman Compose isn't installed anymore.

Since I'm using SteamOS, every system update would remove Podman Compose.

But Winboat thinks I probably still have it? 

I know I don't have it anymore because I deleted my VM to start a new one and now I can see that I don't have Podman Compose anymore:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/704d679e-007b-41f1-a775-fb8d0d788bfb" />

So the issue is that Winboat doesn't tell you this. It needs to tell you if you don't have it installed anymore, because now I deleted my VM for no reason!

ShaunRoselt:
nvm.

It's doing it again and I definitely still have Podman Composer installed.

I spent a whole day setting up my VM yesterday, and now today, it's broken again. I can't get into it. It's just loading forever and nothing actually happens:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0b4cb8d1-410e-4072-9aae-fc2507cb2fbe" />

---

When I delete my VM, then I'm able to make a new one and that one works. I just did it again.

ShaunRoselt:
In case it matters. This is what I'm doing in my VM:

1. Install [Git](https://git-scm.com/)
2. Install [TortoiseGit](https://tortoisegit.org/)
3. Install [RAD Studio 13 Enterprise](https://www.embarcadero.com/products/rad-studio/)
4. Sign into OneDrive
[...]

Issue #569, opened 2025-12-13
[Bug] failed to fulfil mount request: open /dev/bus/usb/001/011: no such file or directory


kodenamekrak:
### What version of WinBoat are you running?

0.9.0 Beta Prod

### Your Environment

CachyOS
KDE
freerdp 3.19.1

### Steps to Reproduce / Context

plugging in my Oculus Quest 2 to my pc while the container is running (for use unrelated to winboat), then presumably shutting down the container with the headset still plugged in

### Logs

[winboat.log](https://github.com/user-attachments/files/24143075/winboat.log)

[container.log](https://github.com/user-attachments/files/24143078/container.log)

### Expected Behavior

Winboat launches as normal 

### Current Behavior

winboat container now refuses to launch when the headset is not plugged in as it is trying to pass through the headset

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

BinaryQuantumSoul:
Same issue on NixOS + Hyprland
Container unable to launch without usb stick

kodenamekrak:
as a workaround i've edited the compose.yml file in `~/.winboat` to only specify the usb devices i actually need (virtualisation, keyboard, mouse)
```yml
    devices:
      - /dev/vhost-net
      - /dev/kvm
      - /dev/bus/usb/001/004
      - /dev/bus/usb/001/006
```

BinaryQuantumSoul:
Possible workaround: https://github.com/TibixDev/winboat/issues/591#issuecomment-3832245283

nkay08:
Hi, could you check #809 and see if it fixes your issue?

Also see #533 and add these to volumes in the `podman-compose.yml`(which is in either `~/.winboat` or for latest dev build in `~/.local/share/winboat-app`:

```
    volumes:
      ...
      - /dev:/dev:rslave
      - /dev/pts:/dev/pts
```

Issue #502, opened 2025-11-20
[Bug] Podman and Winboat can't start the container after i removed my keyboard and mouse usb wireless dongle


ahmed-abdelrazek:
### What version of WinBoat are you running?

v0.9.0-alpha2

### Your Environment

Distro: bazzite stable Linux 6.17.7-ba01.fc43.x86_64 (uptodate i guess fedora kinoite 43)
DE: KDE Plasma 6.5.1
Freerdp: flatpak 3.18.0
Additional info: using the newly supported podman v1.22.1 and appimage winboat
the device is ThinkPad L15 Gen 1 (20U7000UUS)
windows is also uptodate

### Steps to Reproduce / Context

plug a keyboard wireless 2.4ghz usb dongle
install windows
shutdown the pc
removed the usb dongle
try to start winboat again

### Logs

[winboat.log](https://github.com/user-attachments/files/23660363/winboat.log)

[install.log](https://github.com/user-attachments/files/23660366/install.log)

[container.log](https://github.com/user-attachments/files/23660368/container.log)

### Expected Behavior

it should just work no matter what usb device is connected or not connected

### Current Behavior

the poadman pod and containers for winboat refuses to run

<img width="1851" height="218" alt="Image" src="https://github.com/user-attachments/assets/a91cda8a-6e1f-4f89-99d4-033252367126" />

### Quality Notice

- [x] I have checked the issue tracker and verified that this bug is a unique case.

Levev:
This issue got fixed in 8eb44ef. 
We won't support USB passthrough on podman in the 0.9.0 release, as it requires additional permissions.
That being said, we already know how we'll go about implementing USB passthough, and it'll arrive later down the line

PR #809, opened 2026-07-04
Fix: Recreate container from compose file if removed or running into errors

nkay08:
Hi,

I had two issues with Winboat when using it on my steam deck. I am using `podman` and `podman-compose` installed via `brew` which works fine.

However, sometimes the podman container gets broken due to the USB passthrough (e.g. when USB devices have been unplugged in the meantime).
When then trying to start the container in Winboat, there is just an endless spinner without error message.
But manually trying to start it shows this:
```
(deck@steamdeck ~)$ podman ps -a
CONTAINER ID  IMAGE                        COMMAND     CREATED     STATUS                 PORTS                                                                                                                                  NAMES
06cdeee91076  ghcr.io/dockur/windows:5.14              4 days ago  Exited (0) 4 days ago  127.0.0.1:39203->3389/tcp, 127.0.0.1:38997->7148/tcp, 127.0.0.1:37031->7149/tcp, 127.0.0.1:39639->8006/tcp, 127.0.0.1:37213->3389/udp  WinBoat

(deck@steamdeck ~)$ podman start 06cdeee91076
Error: unable to start container "06cdeee9107683c1f2c645dca82cd204d08a9c8d299df2a7c0cf98543d3ecdc5": crun: cannot stat `/dev/bus/usb/001/022`: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

Additionally, if the container was removed for some reason, Winboat will not detect this on start and show the initial setup screen on start, even though all the files of the windows installation still exist.  
When manually starting the container via the compose file, everything works fine again.

So this change includes the following:

1. Detect existing installation files and don't show initial setup screen if there is currently no existing container.
2. Recreate the container if it runs into errors due to USB device mapping
3. Recreate the container from the compose file if installation files exist
4. Don't show endless spinner for some actions if they fail
5. Show error message in GUI if starting of the container fails even after trying to recreate it

These changes are related to:

- #663 
- #569 
- #502
- #591 
- #218
- #533 (<- this change might also be sensible for podman)

### DIFF
--- a/src/renderer/lib/constants.ts
+++ b/src/renderer/lib/constants.ts
@@ -8,6 +8,8 @@ export const WINBOAT_DIR = process.env.XDG_DATA_HOME ?
     path.join(os.homedir(), ".local", "share", "winboat-app");
 export const DEFAULT_HOMEBREW_DIR = path.join(os.homedir(), "../linuxbrew/.linuxbrew/bin");
 
+export const CONTAINER_LOG_FILE = path.join(WINBOAT_DIR, "container.log");
+
 export const WINDOWS_VERSIONS = {
     "11": "Windows 11 Pro",
     "11l": "Windows 11 LTSC 2024",
--- a/src/renderer/lib/containers/container.ts
+++ b/src/renderer/lib/containers/container.ts
@@ -1,10 +1,10 @@
 import { ComposeConfig } from "../../../types";
-import { WINBOAT_DIR } from "../constants";
+import { CONTAINER_LOG_FILE } from "../constants";
 import { createLogger } from "../../utils/log";
 
 const path: typeof import("node:path") = require("node:path");
 
-export const containerLogger = createLogger(path.join(WINBOAT_DIR, "container.log"));
+export const containerLogger = createLogger(CONTAINER_LOG_FILE);
 
 export type ComposeDirection = "up" | "down";
 export type ComposeArguments = "--no-start";
--- a/src/renderer/lib/containers/container.ts
+++ b/src/renderer/lib/containers/container.ts
@@ -36,6 +36,7 @@ export enum ContainerStatus {
     PAUSED = "Paused",
     EXITED = "Exited",
     UNKNOWN = "Unknown",
+    ERROR = "ERROR",
 }
 
 // Errors which usually indicate that the container is in a stale/broken state,
--- a/src/renderer/lib/containers/container.ts
+++ b/src/renderer/lib/containers/container.ts
@@ -37,3 +37,35 @@ export enum ContainerStatus {
     EXITED = "Exited",
     UNKNOWN = "Unknown",
 }
+
+// Errors which usually indicate that the container is in a stale/broken state,
+// e.g. because it references a passed-through USB device that is no longer
+// present on the host. In these cases the container can't be started again and
+// needs to be recreated from the compose file
+const STALE_CONTAINER_ERROR_PATTERNS = [
+    /cannot stat `[^`]*`:?\s*no such file or directory/i,
+    /oci runtime attempted to invoke a command that was not found/i,
+    /no such device or address/i,
+];
+
+function getErrorText(error: unknown): string {
+    if (!error) return "";
+    if (typeof error === "string") return error;
+
+    if (typeof error === "object") {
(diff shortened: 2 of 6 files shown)


waffles-dev:
Sorry I've not been on top of things.
I like the PR but I'm just raising it to Tibix just to double check a couple of things, hoping we can get it in soon.

TibixDev:
Hey, just a couple of things before we merge:
- Can you show me how the last container error UI would look visually?
- How would recreating the container help with stale USB device mappings? Is it an issue on Podman's side that they cache the specific mappings even if the entire bus is passed, when you start it, it tries to use said mapping? At least that's the idea that I get from this fix

Other than that good changes, for example I think the guard clauses are a great change so we don't show a loading spinner forever, thanks!

nkay08:
> Hey, just a couple of things before we merge:
> 
>     * Can you show me how the last container error UI would look visually?
> 
>     * How would recreating the container help with stale USB device mappings? Is it an issue on Podman's side that they cache the specific mappings even if the entire bus is passed, when you start it, it tries to use said mapping? At least that's the idea that I get from this fix
> 
> 
> Other than that good changes, for example I think the guard clauses are a great change so we don't show a loading spinner forever, thanks!

I will try to provide a screenshot of the UI in case of such an error.

Yes, it seems like on podman's side it caches the mappings and fails to start the container again if those have changed.
The error message is similar to this:
`2026-01-31 07:11:23 | ERROR | Error: Command failed: podman container start WinBoat
Error: unable to start container "ff3b71d6282d8d8ccb951c94583f29053fa68ffc69b0d098c5d0195027f74ef2": crun: cannot stat /dev/bus/usb/001/004: No such file or directory: OCI runtime attempted to invoke a command that was not found`
I also tried out another PR [#553](https://github.com/TibixDev/winboat/pull/533)  which suggests some changes to the compose file and this actually solves this specific problem.

I think both PRs are sensible. PR [#553](https://github.com/TibixDev/winboat/pull/533) alleviates this specific podman issue, and my PR introduces container recreation as a fallback.

nkay08:
This is how the UI for the container error looks like.

winboat_container_error

TibixDev:
Do you think you can make it look a little bit nicer and make it have a clickable text for "container logs" that opens up the actual container logs?

nkay08:
> Do you think you can make it look a little bit nicer and make it have a clickable text for "container logs" that opens up the actual container logs?

I can add the function to open the container logs, sure.  

To be honest, I am not really sure what defines a "nicer" look (aesthetics is not really my proficiency).  
What did you have in your mind?  

In the best case this will hopefully be rarely shown to the user.

nkay08:
winboat_error_message

Is this maybe more how you imagined it to look like?
(haven't committed the visual changes yet)

waffles-dev:
Ah - sorry to be annoying, but instead of adding the new box, can you try updating the `Container - Unknown` text to `Container - Failed to Start (check logs)` (where logs is the link)?

If it can fit in the one, pre-existing line, it'd be ideal so that we're not pushing the vertical space of that element out.

nkay08:
winboat_error_message_inline

Ok got it, so how about this?

I introduced a new `ContainerStatus.ERROR` in my last commit to make the the condition logic in `Home.vue` easier, and thus removed the newly added `Winboat.lastContainerError` again.

(Instead of "Failed to start" it could also just show "Error" like for the other statuses)


P.S. I did a rebase and was initially confused why the added installation detection no longer worked. I very much appreciate to move the `WINBOAT_DIR` directory to `XDG_DATA_HOME`. Maybe it is already planned, but could there maybe be an automatic migration from the old location or a message for users that update?

waffles-dev:
Nice - I'll get Tibix to do a final check of this one.
re. the migration, not sure yet. Might be doing a breaking change for a 1.0

TibixDev:
As for migration: Not quite sure about 1.0 being backwards compatible, probably not.

The UI looks good now if you ask me, and if the fix works for Podman users than all the better. If possible, in all these linked issues I'd like you to try and ask them if this fix remediates the problem and if yes perhaps leave a comment here with all the issues that can be closed. I'll eventually look back myself but it's faster if someone dedicated to this does it.

Thank you! 🙌

PR #813, opened 2026-07-12
feat: Allow NAT when running with Podman

kroese:
I think the line

```yaml
  NETWORK: "user"
```

was added in the past to workaround some issue that does not exist any longer in v6.00 of the image.

I would remove it, so that Podman users can use NAT (and silently fallback to usermode when needed).

Unless there is a very clear reason why usermode was forced.


TibixDev:
LGTM, Merging as per #818 🙌

PR #814, opened 2026-07-13
fix: Replace profile picture with local file

LittleT6109:
The link used was no longer working, so I put the image locally so we never have to worry about that again.


LittleT6109:
Had to force push a few times to fix some messy history, I merged main into the branch on GitHub instead of rebasing locally.

TibixDev:
LGTM! 🙌

PR #816, opened 2026-07-13
feat: Respect XDG_DATA_HOME for storing app-related files instead of ~/.winboat

### LINKED ISSUE #77: Please store data in XDG directories instead of ~/.winboat
There is enough software that already clutter home folder with dot files. 

`$XDG_CACHE_HOME` for cache (default: `~/.cache`)
`$XDG_DATA_HOME` for data (default: `~/.local/share`)
`$XDG_CONFIG_HOME` for configs that for user to edit (default: `~/.config`)
`$XDG_STATE_HOME` for history and logs (default: `~/.local/state`)

But even dumping everything in ~/.local/share or ~/.config is good enough.

TibixDev:
This PR aims to change the `~/.winboat` directory for app related files to `$XDG_DATA_HOME/winboat-app`. While not a 100% perfect solution (we could split app related files into many more XDG dirs), it respects the spec enough that most people should be happy, or at least happier than having a dot folder in their home.

❗Needs migration code for existing users if we decide that 1.0 will be backwards compatible

Closes #77

### DIFF
--- a/src/renderer/lib/config.ts
+++ b/src/renderer/lib/config.ts
@@ -153,7 +153,7 @@ export class WinboatConfig {
             if (!writeDefault) return null;
             // Also the create the directory because we're not guaranteed to have it
             if (!fs.existsSync(WINBOAT_DIR)) {
-                fs.mkdirSync(WINBOAT_DIR);
+                fs.mkdirSync(WINBOAT_DIR, { recursive: true });
             }
 
             fs.writeFileSync(WinboatConfig.configPath, JSON.stringify(defaultConfig, null, 4), "utf-8");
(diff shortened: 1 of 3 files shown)


waffles-dev (approved):
lgtm - if it's safe to assume .local/share will always exist on every OS - I think we just use a `mkdir` not a `mkdir -p` when we go to create, so that could potentially be an issue - but if they don't have .local/share they probably have bigger issues.

TibixDev:
@waffles-dev Haha yeah agreed, regardless I added `{ recursive: true }` (`-p` equivalent) to be 100% sure

PR #817, opened 2026-07-13
refactor: Remove all dynamic port related code, features, settings, conditions

TibixDev:
This PR aims to remove dynamic ports. I've concluded that it doesn't make much sense to have dynamic ports in WinBoat for the following reasons:
- Massive technical debt, not only in code size (526 SLOC removed) but also having to watch for initialization order in many places, harder to reason-about logic, and more verbosity
- It makes things much harder to debug, not only on the developer side but also on the user side, always requiring an extra step to find the currently allocated port(s) before any substantial debugging can be done
- Almost nobody used the feature to change ports, at least not that I know of. On the other hand there were plenty people who complained about dynamic ports
- Docker often (always?) allocates a port one step higher in the range upon container restarts, for no apparent reasons
- Podman doesn't even support dynamic port ranges, instead you are the mercy of getting a random port
- Many apps have their own dedicated ports that they expect to be free, I feel like it's not a hard demand for us to be accomodated in a 65k port range

WinBoat from now on will reserve the following ports:
```
HOST_NOVNC_PORT = 47270;
HOST_API_PORT = 47271;
HOST_QMP_PORT = 47272;
HOST_RDP_PORT = 47273;
```
With the expectation that we will not surpass `47280` even in the future as we might need more ports. From my checks I've established no app currently that uses this port range by default according to the IANA registry.

❗Needs migration code for existing users if we decide that 1.0 will be backwards compatible


waffles-dev (approved):
LGTM

PR #819, opened 2026-07-15
docs: IP tables documentation

kroese:
The current wording is incorrect, it makes it seems as if NAT only increases LAN performance, but the gain is equally large for WAN performance. So I simplified the text a bit so it reflects the real advantage better.


TibixDev:
Thank you for correcting this, if you want you can also submit an edit here:
https://www.winboat.app/docs/help/firewall-config/

(Just noticed the link wording is wrong, @LittleT6109 if you could take a look)

LittleT6109:
@TibixDev https://github.com/TibixDev/winboat-web/pull/10

Also what do you mean by link wording being wrong? If you are referring to the www. subdomain, both seem to load the page. Starlight automatically changes the links to lowercase, so `Firewall-config.md` becomes `firewall-config`

PR #821, opened 2026-07-17
Update Windows Docker image version to v6.03

kroese:
Update Windows Docker image version to v6.02

### DIFF
--- a/src/renderer/data/docker.ts
+++ b/src/renderer/data/docker.ts
@@ -14,7 +14,7 @@ export const DOCKER_DEFAULT_COMPOSE: ComposeConfig = {
     },
     services: {
         windows: {
-            image: "ghcr.io/dockur/windows:6.02",
+            image: "ghcr.io/dockur/windows:6.03",
             container_name: "WinBoat",
             environment: {
                 VERSION: "11",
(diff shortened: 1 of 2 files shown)


TibixDev:
It seems like this also fixes the Podman port problem for good, thanks @kroese  :)

kroese:
@TibixDev Yes this release contains many Podman networking-related fixes. Especially for rootless Podman (which are less relevant because you always set `privileged: true`).

I have two suggestions:

   - I really would like to see WinBoat get rid of the `privileged: true` flag for both Docker and Podman. We discussed this before and you said that was not possible because of the way you share the home folder (?). I cannot recall the specifics anymore. So maybe a start would be to only inject `privileged: true` when the user actually enables the folder-sharing feature. That way it never runs needlessly with too much privileges, and we also have a nice way to disable `privileged` from the UI for testing purposes.

  - Secondly, you explained that you were very busy and didnt have time to work on this project. But is it possible that you just build WinBoat releases also when I release a new version? So if I release a new Docker image that ideally would trigger also a new WinBoat release. And even though WinBoat stays the same, it still is beneficial for users since it includes a new version of `dockur/windows` so it still is an improvement. Another alternative would be that you set the image tag to `:latest`. That way you dont have to create new WinBoat releases. But tagging the image would be preferred ofcourse.

PR #823, opened 2026-07-26
feat: New guest server & guest server updater model

TibixDev:
This PR changes how the guest server (GS) and the guest server updater (GSU) works. There's now a new separate native service whose sole responsibility is updating the GS, instead of the previous approach where the GS attempted to update itself through a hacky Powershell script (the more of these scripts we can get rid of, the better).

We also got rid of the password hashes, there's now a single shared token for authentication. If an attacker has read/write access on the machine, they can compromise the guest server either way without much effort. It was largely unnecessary code.

The new GSU does the update in a safe way where it can also easily roll back if something goes wrong.


TibixDev:
Also, pinned Jimp because the new version is broken at the moment, it's an issue on their end.

PR #830, opened 2026-08-06
fix: configure NVIDIA Helios containers

rupansh:
Map render nodes to exact NVIDIA UUIDs, require runtime or CDI support, and keep Vulkan, EGL, and GBM on the NVIDIA provider.

depends on https://github.com/winboat-org/helios-windows/pull/1


TibixDev:
LGTM! 🚀

PR #833, opened 2026-08-08
Update Windows Docker image version to v6.05

kroese:
Updates base image to v6.05

This release contains more changes than ever, but the most noticeable will be that it does not need to rebuild the .iso file anymore. That skips the whole extracting/rebuilding phases after download, so it goes straight from download to booting the VM, which saves a lot of time (especially on slow storage).


kroese:
@TibixDev Can this be included in 0.9.1 please

PR #848, opened 2026-08-18
fix: Recognize all auto-restart policies for Auto Start Container

### LINKED ISSUE #474: [Bug] Auto Start Container remains active

0.8.7 Prod

### Your Environment

Distro: Debian Trixie
DE: Gnome

The winBoat guest keeps autostarting. Deactivating "Auto Start Container" in the Configuration and clicking save does not stick, when switching to "Home" and back to "Configuration", it's active again.

### Logs

[winboat.log](https://github.com/user-attachments/files/23496046/winboat.log)

[install.log](https://github.com/user-attachments/files/23496053/install.log)

### Expected Behavior

Configuration should change, container should not autostart.

### Current Behavior

Configuration change does not work, container keeps autostarting.

### Possible Solution

Accept configuration change, do not autostart.

### Quality Notice

Pager-dot:
The Auto Start Container switch read the compose restart policy with a strict equality check against `unless-stopped`, so the two other policies that also bring the container back on their own (`always` and the legacy `on-failure` default) rendered as OFF.

Installs created before 4fefd6d shipped `restart: on-failure`. For those users the switch showed OFF while Docker still restarted the container, and because `origAutoStartContainer` matched, the Save button stayed disabled - leaving no in-app way to write `restart: no`.

Read the policy against a shared AUTOSTART_RESTART_POLICIES list instead, and track whether the on-disk policy is non-canonical so Save stays reachable and one save normalizes the file to `unless-stopped` or `no`.

Fixes #474

### DIFF
--- a/src/renderer/lib/constants.ts
+++ b/src/renderer/lib/constants.ts
@@ -111,3 +111,14 @@ export const RESTART_UNLESS_STOPPED = "unless-stopped";
 export const RESTART_ON_FAILURE = "on-failure";
 export const RESTART_ALWAYS = "always";
 export const RESTART_NO = "no";
+
+/**
+ * Restart policies under which the container comes back up on its own.
+ * Legacy installs shipped `on-failure` as the default, so it has to be
+ * recognized here even though we never write it anymore.
+ */
+export const AUTOSTART_RESTART_POLICIES: string[] = [
+    RESTART_UNLESS_STOPPED,
+    RESTART_ALWAYS,
+    RESTART_ON_FAILURE,
+];
--- a/src/renderer/lib/constants.ts
+++ b/src/renderer/lib/constants.ts
@@ -112,13 +112,5 @@ export const RESTART_ON_FAILURE = "on-failure";
 export const RESTART_ALWAYS = "always";
 export const RESTART_NO = "no";
 
-/**
- * Restart policies under which the container comes back up on its own.
- * Legacy installs shipped `on-failure` as the default, so it has to be
- * recognized here even though we never write it anymore.
- */
-export const AUTOSTART_RESTART_POLICIES: string[] = [
-    RESTART_UNLESS_STOPPED,
-    RESTART_ALWAYS,
-    RESTART_ON_FAILURE,
-];
+/** Restart policies under which the container comes back up on its own */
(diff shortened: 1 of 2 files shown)


src/renderer/lib/constants.ts
```diff
@@ -111,3 +111,6 @@ export const RESTART_UNLESS_STOPPED = "unless-stopped";
 export const RESTART_ON_FAILURE = "on-failure";
 export const RESTART_ALWAYS = "always";
 export const RESTART_NO = "no";
+
+/** Restart policies under which the container comes back up on its own */
+export const AUTOSTART_RESTART_POLICIES: string[] = [RESTART_UNLESS_STOPPED, RESTART_ALWAYS];
```
waffles-dev, line 116: should we keep ON_FAILURE in here in case the user adds it manually/externally to the compose?

Pager-dot: I don't think we should — `on-failure` doesn't actually auto-start on daemon restart.

Docker's docs are explicit about this:

> The `on-failure` policy only prompts a restart if the container exits with a failure. **It doesn't restart the container if the daemon restarts.**

See the [`docker container run` reference](https://docs.docker.com/reference/cli/docker/container/run/) and [Start containers automatically](https://docs.docker.com/engine/containers/start-containers-automatically/). The latter also makes the distinction clear: only `always` and `unless-stopped` restart containers when the Docker daemon starts.

So for a host reboot, an `on-failure` container stays down. It only recovers from a non-zero exit while the daemon is already running — that's crash recovery rather than auto-start.

Since this toggle is specifically described as **"automatically be started when the system boots up"**, I think showing `on-failure` as OFF is the accurate behavior, including when a user has configured it manually. Their existing restart policy is also left untouched unless they explicitly change the toggle.

This also means my PR description's framing around the legacy `on-failure` default silently auto-starting was incorrect. Those containers wouldn't have come back after a host reboot either. The actual issue was that the toggle showed OFF while Save was unreachable, which is what this change fixes. I'll correct that part of the description.

I've also been running the patched build locally, and the toggle round-trips correctly in both directions:

- **Enable** → `unless-stopped`
- **Disable** → `no`

The recreated container reflects each change (`docker inspect` confirms it), and the Compose backups show the expected transitions.

So I'd suggest keeping this as:

```ts
[RESTART_UNLESS_STOPPED, RESTART_ALWAYS]

PR #881, opened 2026-09-18
fix(setup): check for HeliosSetup.exe instead of stale Install-Helios.ps1

idiazraz:
## Summary
- The Setup wizard's `heliosAvailable` check (`src/renderer/views/SetupUI.vue`) still looks for `helios/Install-Helios.ps1`, a path that stopped being shipped once the Helios GPU bundle moved to a self-contained `HeliosSetup.exe` installer (b893be7, f4cb638).
- `install.ts`'s own preflight check was updated to match at the time (`path.join(appPath, "helios", "HeliosSetup.exe")`), but this UI gate was missed.
- Net effect: on any build using a current Helios bundle, the GPU acceleration checkbox in the setup wizard is permanently disabled with "This WinBoat build does not include the Helios driver bundle" — even though the driver is present and `install.ts` would install it without issue.
- Fix: check for `HeliosSetup.exe` in `SetupUI.vue`, matching `install.ts`.

Found this while building `1.0-bleeding-edge` from source with an AMD GPU (`amdgpu`, non-NVIDIA render-node path) — the wizard blocked GPU acceleration despite a correctly built Helios bundle in `guest_server/dist/oem/helios/`.

## Test plan
- [x] Built the guest server via `build-guest-server.sh`, confirmed `guest_server/dist/oem/helios/HeliosSetup.exe` exists (no `Install-Helios.ps1` in the current bundle)
- [x] Rebuilt the Electron app with the fix; Setup wizard now shows the GPU acceleration checkbox enabled with render devices listed instead of the "does not include the Helios driver bundle" error


idiazraz:
Closing for now while I look into a more substantial GPU acceleration contribution — may reopen or fold this fix into a larger change later.

Djihads80:
this fixed my issue

best to put it on hold then outright close it no?

